PR #1, opened 2023-04-13
Upgrade the converter

Zaid-Ajaj:
- Using latest pulumi (master) to include https://github.com/pulumi/pulumi/pull/12661
- Using latest pulumi-terraform-bridge to include https://github.com/pulumi/pulumi-terraform-bridge/pull/982
- `tfconvert.TranslateModule` now needs a `sourceDirectory: string` which I've provided as `"/"`
- `convert.NewMapper(req.MapperTarget)` is no longer a thing, changed to `convert.NewMapperClient(req.MapperTarget)` instead


cmd/pulumi-converter-terraform/main.go
```diff
@@ -51,7 +51,7 @@ func (*tfConverter) ConvertProgram(ctx context.Context,
 	src := afero.NewBasePathFs(fs, req.SourceDirectory)
```
Frassle, line 51:
```suggestion
```
Delete this

cmd/pulumi-converter-terraform/main.go
```diff
@@ -51,7 +51,7 @@ func (*tfConverter) ConvertProgram(ctx context.Context,
...
 	dst := afero.NewBasePathFs(fs, req.TargetDirectory)
 
-	diags := tfconvert.TranslateModule(src, dst, providerInfoSource)
+	diags := tfconvert.TranslateModule(src, "/", dst, providerInfoSource)
```
Frassle, line 54:
```suggestion
	diags := tfconvert.TranslateModule(fs, req.SourceDirectory, dst, providerInfoSource)
```
Give TranslateModule full access to the file system to allow translation of modules that are using relative paths above the root module.

Frassle (approved):

PR #22, opened 2023-08-22
Add example test suite

### LINKED ISSUE #5: Add test/validation framework
Test bed of "medium-sized" Terraform codebases

justinvp:
Adds example test suite that runs for PRs and in a cron job.

Fixes #5


.github/workflows/stage-test.yml
```diff
@@ -38,20 +38,54 @@ jobs:
...
+
+      # Pre-install some needed plugins
+      - name: Install pulumi plugins
         run: |
-          if [ "${{ matrix.go-version }}" != "1.17.x" ]; then
-            go mod tidy -compat=1.17
-          fi
-      - name: Build
-        run: make build
+          pulumi plugin install resource aws
```
Zaid-Ajaj, line 62: Probably a good idea to pin those to a specific major version

tests/example_test.go
```diff
@@ -0,0 +1,494 @@
...
+	for k := range ss {
+		if !other.Has(k) {
+			return false
+		}
+	}
+	return true
+}
+
+const (
+	csharp     = "c#"
```
Zaid-Ajaj, line 98: I don't believe we don't use the literal `c#` as a source language, instead `pulumi convert` accepts `csharp` or `dotnet`

Zaid-Ajaj: Nevermind actually, it was just my terminal that didn't like `c#` but it worked when I used "c#" 😅

Zaid-Ajaj (approved):
Very nice to have this! LGTM

Issue #36, opened 2023-09-29
Enable a bulk interface for converting examples


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

For bridged providers that do example discovery and conversion from TF HCL to PCL and further on to TypeScript and other languages, it would be great to have a command-line interface to convert many (thousands) of examples at once.

This is partially addressed
https://github.com/pulumi/pulumi-converter-terraform/pull/29

However as written #29 does not allow the caller to set the input file name so the functionality cannot be used. 

Suggested alteration is to accept it in PULUMI_CONVERT_EXAMPLES environment variable or use some form of by-convention "examples.json" processing.

t0yv0:
This works now with #29 latest changes.

PR #37, opened 2023-09-29
Use an env var for examples.json location

t0yv0:
I could never get it to work with plain CLI arguments.

Fixes https://github.com/pulumi/pulumi-converter-terraform/issues/36

Otherwise it works well.


t0yv0:
Closing in favor of https://github.com/pulumi/pulumi-converter-terraform/pull/29 which works now! Great!

PR #73, opened 2023-10-13
Parallelize the example conversion

t0yv0:
Trying this out as AWS build is a little sluggish. This increases CPU utilization in the converter  by loading all the cores and speeds things up a little. This is part of the solution that speeds up AWS from 30-40 min to <7 min, highly appreciated to improve iteration speed.

### DIFF
--- a/cmd/pulumi-converter-terraform/par.go
+++ b/cmd/pulumi-converter-terraform/par.go
@@ -1,0 +1,126 @@
+// Copyright 2016-2023, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"runtime"
+	"sync"
+)
+
+// Transforms a large map in batches of up to batch elements, using workers number of goroutines. If
+// workers is -1, uses one worker per CPU.
+func parTransformMap[K comparable, T any, U any](
+	inputs map[K]T,
+	transform func(map[K]T) (map[K]U, error),
+	workers int,
+	batch int,
+) (map[K]U, error) {
+	if batch < 1 {
+		return nil, fmt.Errorf("batch cannot be less than 1")
+	}
+	n := workers
+	if workers < 1 {
+		n = runtime.NumCPU()
+		if n < 2 {
+			n = 2
+		}
+	}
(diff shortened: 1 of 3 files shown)


cmd/pulumi-converter-terraform/par.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	translatedMap := map[K]U{}
+	for _, k := range keys {
+		translatedMap[k] = translations[keyIndex[k]]
+	}
+	return translatedMap, nil
+}
+
+func parTransformMapWith[K comparable, T any, U any](
```
Frassle, line 108: Feels like you could simplify things if you just had this version of the function. No need for the func(map) version as well.

cmd/pulumi-converter-terraform/par.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	n := workers
+	if workers < 1 {
+		n = runtime.NumCPU()
+		if n < 2 {
+			n = 2
+		}
+	}
+
+	keys := []K{}
```
Frassle, line 42: I don't think you need a slice here. Just have three+ goroutines spinning.
input) a goroutine iterating the input map and feeding a channel with (key, value) pairs
N workers) reading from that channel and running the mapping function and writing to another (key, value) channel
output) reading from the channel workers is writing to and building a new map off those pairs

No temporary maps, no slice counting, and full use of all worker till everything is translated.

cmd/pulumi-converter-terraform/par.go
t0yv0: Yeah. Copy-pasting from another context here the other function version made more sense.

cmd/pulumi-converter-terraform/par.go
t0yv0: Eh, sure.

cmd/pulumi-converter-terraform/par.go
```diff
@@ -0,0 +1,87 @@
...
+// limitations under the License.
+
+package main
+
+import (
+	"runtime"
+	"sync"
+)
+
+// Transforms map values in parallel over n workers. If workers is negaive use NumCPU.
```
Frassle, line 22:
```suggestion
// Transforms map values in parallel over n workers. If workers is negative use NumCPU.
```

cmd/pulumi-converter-terraform/par.go
```diff
@@ -0,0 +1,87 @@
...
+	for k, v := range inputs {
+		ch <- kv{k, v}
+	}
+
+	close(ch)
+
+	wg.Wait()
+
+	for _, e := range errors {
+		// Returning the first error here, could instead consider merging them.
```
Frassle, line 74: Rename `errors` to `errs` so you can use the errors package then could just do something like:
```
err := errors.Join(errs...)
if err != nil { return nil, err }
```

t0yv0: Sure why not. It's not a clean win because it breaks symmetry between which errors are returned from regular and parallel code, but clearly that doesn't matter too much in here.

Frassle (approved):

PR #77, opened 2023-10-20
Test the diagnostics we generate as well as the code

Frassle:
The diagnostics we emit from the converter are just as important as the code. This adds snapshot testing of diagnostics for our test programs.


iwahbe (approved):
LGTM

Frassle:
Hmm some non determinism in diagnostic order, will look into that.

Frassle:
Was just filenames not being relative, fixed in the test.

PR #85, opened 2023-11-14
Fix order of parameters for the `parseint` function

Zaid-Ajaj:
Fixes https://github.com/pulumi/pulumi/issues/14554


Frassle (approved):
Total facepalm that we didn't notice this looking through the pcl/main.pp file before.

tgummerer (approved):

PR #97, opened 2024-01-16
add remaining std functions

dixler:
This PR adds the remaining functions implemented in `pulumi-std`.


Frassle:
Most of these are "generic" and won't ever be added to `pulumi-std`, the few that are already in std are due to be removed.

PR #101, opened 2024-01-29
Test that map keys don't get renamed

Frassle:
Fixes https://github.com/pulumi/pulumi-converter-terraform/issues/100.

This doesn't fix every rename, there's still cases where we should be leaving keys as is but are renaming them.

### DIFF
--- a/pkg/convert/testdata/programs/map_keys/pcl/main.pp
+++ b/pkg/convert/testdata/programs/map_keys/pcl/main.pp
@@ -1,0 +1,18 @@
+resource "aResource" "complex:index/index:resource" {
+  __logicalName = "a_resource"
+  aMapOfBool = {
+    CAPS       = true
+    camelCase  = false
+    snake_case = false
+    PascalCase = true
+  }
+}
+
+output "someMapOutput" {
+  value = {
+    CAPS       = aResource.result
+    camelCase  = 1
+    snakeCase  = 2
+    pascalCase = 3
+  }
+}
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -2,6 +2,8 @@
 
   - Support the `replace` function.
 
+  - Don't rename map keys.
+
 ### Bug Fixes
 
  - Fix order of parameters for the `parseint` function
--- a/pkg/convert/testdata/programs/map_keys/main.tf
+++ b/pkg/convert/testdata/programs/map_keys/main.tf
@@ -1,0 +1,17 @@
+resource "complex_resource" "a_resource" {
+    a_map_of_bool = {
+        CAPS: true
+        camelCase: false
+        snake_case: false
+        PascalCase: true
+    }
+}
+
+output "some_map_output" {
+    value = {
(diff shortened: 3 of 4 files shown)


pkg/convert/testdata/programs/map_keys/pcl/main.pp
```diff
@@ -0,0 +1,18 @@
...
+    snake_case = false
+    PascalCase = true
+  }
+}
+
+output "someMapOutput" {
+  value = {
+    CAPS       = aResource.result
+    camelCase  = 1
+    snakeCase  = 2
```
Frassle, line 15: Note that these are still being renamed. Fixing this is a bit more involved. If we just flip the logic to assume things are maps not objects then we end up renaming a lot of `{ k: v}` expressions in TF source code which are then used for object values.

Frassle:
Actually this probably fixes https://github.com/pulumi/pulumi-converter-terraform/issues/53 as well.

justinvp:
Sorry, I need to take care of https://github.com/pulumi/pulumi-converter-terraform/issues/99

pkg/convert/testdata/programs/map_keys/pcl/main.pp
Zaid-Ajaj: It is probably fine that output objects _are_ renamed because the type of these becomes are object type from component outputs, not a map.

pkg/convert/testdata/programs/registry_module/pcl/main.pp
```diff
@@ -1,11 +1,11 @@
...
   networks = [{
-    name    = "foo"
-    newBits = 8
+    name     = "foo"
+    new_bits = 8
     }, {
-    name    = "bar"
-    newBits = 8
+    name     = "bar"
+    new_bits = 8
```
Zaid-Ajaj, line 8: Is networks here a list of maps?

Frassle: https://registry.terraform.io/modules/hashicorp/subnets/cidr/latest?tab=inputs

`networks list(object({ name = string new_bits = number }))`

object() looks like we should have marked this as _not_ a map. So I'll look into this.

Frassle: Yup so this is two problems
1. I screwed up and flipped the assumption of defaulting to rename if we didn't know the type.
2. We're not actually getting the type for this, even though we _probably_ could. I think we should merge this PR and then open a new one to look at seeing if we can write a module using maps and objects and check that translates correctly.

Zaid-Ajaj: > We're not actually getting the type for this, even though we probably could

We should get the type from the mapping

> we should merge this PR and then open a new one to look at seeing if we can write a module using maps and objects and check that translates correctly.

Sounds good!

justinvp:
Tests are unblocked and should now pass with https://github.com/pulumi/pulumi-converter-terraform/pull/105

Zaid-Ajaj (approved):

PR #107, opened 2024-02-10
Use the type name to help resolve name clashes

Frassle:


pkg/convert/testdata/mappings/simple.json
```diff
@@ -31,6 +31,16 @@
                     "type": 4,
                     "computed": true
                 }
+            },
```
Frassle, line 34: I needed another type with a different name to test this case.

pkg/convert/testdata/programs/name_conflict/pcl/main.pp
```diff
@@ -8,11 +8,16 @@
   inputTwo      = myaThing
 }
 
-aThingData = invoke("simple:index:dataSource", {
+aThingDataSource = invoke("simple:index:dataSource", {
   inputOne = "Hello ${aThingResource.result}"
   inputTwo = myaThing
 })
 
+resource "aThingAnotherResource" "simple:index:anotherResource" {
```
Frassle, line 16: Note that before this would have just ended up being called "aThingResource2"

Zaid-Ajaj (approved):

PR #110, opened 2024-02-14
Sort generated properties by their keys to get deterministic output.

### LINKED ISSUE #109: Non-determinism in HCL->PCL generation
### What happened?

When attempting to use the latest version of the converter in the Azure example rendering pipeline, the converter does not always generate the same PCL for a given HCL example. Unfortunately our builds are very sensitive to that because the generated example projections into Python and TypeScript are spliced into the SDKs and checked into verbatim. The non-determinism makes the builds fail and would require us to rethink the build / check strategy if we were to live with it.

The example diffs are like this:

```
    -         attachedDataNetworkId: testdataazurermMobileNetworkAttachedDataNetwork.id,
    -         sliceId: test.id,
    +         attachedDataNetworkId: test.id,
    +         sliceId: testazurermMobileNetworkSlice.id,
```

Something gets named differently and ordered differently. 

One of the offending examples is like this in the source:

```terraform
resource "azurerm_resource_group" "example" {
  name     = "example-resources"
  location = "West Europe"
}

resource "azurerm_mobile_network" "example" {
  name                = "example-mn"
  location            = azurerm_resource_group.example.location
  resource_group_name = azurerm_resource_group.example.name
  mobile_country_code = "001"
  mobile_network_code = "01"
}

resource "azurerm_mobile_network_sim_group" "example" {
  name              = "example-mnsg"
  location          = azurerm_resource_group.example.location
  mobile_network_id = azurerm_mobile_network.example.id
}

resource "azurerm_mobile_network_slice" "example" {
  name              = "example-slice"
  mobile_network_id = azurerm_mobile_network.example.id
  location          = azurerm_resource_group.example.location

  single_network_slice_selection_assistance_information {
    slice_service_type = 1
  }
}

resource "azurerm_mobile_network_attached_data_network" "example" {
  mobile_network_data_network_name            = azurerm_mobile_network_data_network.example.name
  mobile_network_packet_core_data_plane_id    = azurerm_mobile_network_packet_core_data_plane.example.id
  location                                    = azurerm_resource_group.example.location
  dns_addresses                               = ["1.1.1.1"]
  user_equipment_address_pool_prefixes        = ["2.4.0.0/24"]
  user_equipment_static_address_pool_prefixes = ["2.4.1.0/24"]
  user_plane_access_name                      = "test"
  user_plane_access_ipv4_address              = "10.204.141.4"
  user_plane_access_ipv4_gateway              = "10.204.141.1"
  user_plane_access_ipv4_subnet               = "10.204.141.0/24"
}

resource "azurerm_mobile_network_sim" "example" {
  name                                     = "example-sim"
  mobile_network_sim_group_id              = azurerm_mobile_network_sim_group.example.id
  authentication_key                       = "00000000000000000000000000000000"
  integrated_circuit_card_identifier       = "8900000000000000000"
  international_mobile_subscriber_identity = "000000000000000"
  operator_key_code                        = "00000000000000000000000000000000"

  static_ip_configuration {
    attached_data_network_id = data.azurerm_mobile_network_attached_data_network.test.id
    slice_id                 = azurerm_mobile_network_slice.test.id
    static_ipv4_address      = "2.4.0.1"
  }
}
```

See the attached gist with full repro.

### Example

Isolated to a small repro here: https://gist.github.com/t0yv0/75572d1ef9eed6ac5012754a2a84ab90

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Zaid-Ajaj:
Fixes #109


pkg/convert/tf.go
```diff
@@ -993,6 +993,10 @@ func convertObjectConsExpr(state *convertState, inBlock bool, scopes *scopes,
 			Value: valueTokens,
 		})
 	}
+	// Sort the items by name to make the output deterministic
```
Frassle, line 996: Try and sort by HCL source position. That keeps things in the declaration order from the original.

Zaid-Ajaj: Okay that makes more sense

Zaid-Ajaj: Done ✅

justinvp:
The sorting makes the naming deterministic (what Anton was seeing in #110)?

Zaid-Ajaj:
@Frassle I think I have fixed the non-determinism in naming.  When we rewrite the traversal of invokes/data sources we use `scopes.getOrAddPulumiName` effectively changing the state during traversal without distinguishing between bound vs. unbound invokes. I simplified it so that we don't mutate the scope when we encounter an unbound invoke/data source and instead just renaming it. Same for unbound resources not using `scopes.getOrAddPulumiName` 

The renaming rules for unbound stuff is simple. for data sources `data.{data_source_token}.{local_name}.{rest}` becomes `{localName}{DateSourceToken}.{rest}` 

For unbound resources `{resource_type}.{local_name}.{rest}` becomes `{localName}{resourceType}.{rest}` (similar to what it was doing before)

pkg/convert/tf.go
```diff
@@ -952,6 +952,12 @@ func convertObjectConsExpr(state *convertState, inBlock bool, scopes *scopes,
 	fullyQualifiedPath string, expr *hclsyntax.ObjectConsExpr,
 ) hclwrite.Tokens {
 	items := []hclwrite.ObjectAttrTokens{}
+
+	// Sort the items by line number so that we can deterministically order them in the output
```
Frassle, line 956: I don't think these sorts are actually needed, pretty sure Items is already going to be in declaration order.

pkg/convert/testdata/programs/partial_attribute_lookup/pcl/main.pp
```diff
@@ -1,4 +1,4 @@
 resource "example" "simple:index:resource" {
-  inputOne = exampledataunknownDataSource.attr
-  inputTwo = exampleunknownResourceType.list[0]
+  inputOne = exampleUnknownDataSource.attr
```
Frassle, line 2: These names look better

pkg/convert/tf.go
```diff
@@ -1299,10 +1314,12 @@ func rewriteTraversal(
 					newTraversal = append(newTraversal, hcl.TraverseRoot{Name: newName})
 					newTraversal = append(newTraversal, rewriteRelativeTraversal(scopes, "", traversal[1:])...)
 				} else {
-					// We don't know what this is, so lets assume it's an unknown resource (we shouldn't ever have unknown locals)
-					newName = scopes.getOrAddPulumiName(path, "", camelCaseName(root.Name))
```
Frassle, line 1303: So I think the concern of _not_ using getOrAddPulumiName is that if the picked name here conflicts with some other bound or unbound name that's already in use we'll get a conflict in the PCL.
Feels like we could probably write a test case to cover that, and then fix this up.

Frassle (commented):
I don't think this helps with Antons case at all but the unbound names do look nicer if we can get them without conflicts with other things.

justinvp:
Note: https://github.com/pulumi/pulumi-converter-terraform/pull/116 should address the example test failure.

pkg/convert/tf.go
Frassle: Again I'm pretty sure Items is already sorted, if so this is just extra code for no benefit.

pkg/convert/tf.go
```diff
@@ -1195,9 +1207,19 @@ func rewriteTraversal(
 			// pulumi invoked value instead.
 			suffix := camelCaseName(maybeFirstAttr.Name)
 			path := "data." + maybeFirstAttr.Name + "." + maybeSecondAttr.Name
-			newName := scopes.getOrAddPulumiName(path, "", "data"+suffix)
-			newTraversal = append(newTraversal, hcl.TraverseRoot{Name: newName})
-			newTraversal = append(newTraversal, rewriteRelativeTraversal(scopes, path, traversal[3:])...)
+			if scopes.lookup(path) != "" {
```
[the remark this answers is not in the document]
Zaid-Ajaj: Changed it to use the result from `lookup` ✅ it is technically the same as `camelCaseName(secondAttr.name)` but better get it from the lookup table

pkg/convert/tf.go
```diff
@@ -1589,6 +1614,10 @@ type bodyAttrTokens struct {
 
 func tokensForObject(ts []bodyAttrTokens) hclwrite.Tokens {
 	attrs := make([]hclwrite.ObjectAttrTokens, 0, len(ts))
+	// sort body attribute tokens by name to get a deterministic output
```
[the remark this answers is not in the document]
Zaid-Ajaj: Did these sorts as a just in case because wasn't sure what was and what wasn't sorted. Removed them now

Frassle (approved):

PR #125, opened 2024-03-20
Convert expressions inside `jsonencode` calls without rewriting object keys to camelCase

### LINKED ISSUE #123: aws.iam.Role basic example - invalid policy doc format
https://www.pulumi.com/registry/packages/aws/api-docs/iam/role/#basic-example

The primary example on this page, fails to provision as is. This is due to the JSON policy document in the example needs to have their properties start with uppercase letters, e.g. `statement` should be `Statement`. This goes for all the properties in the policy doc.

Zaid-Ajaj:
### Description

Fixes #123 where we special case `jsonencode` to maintain the casing of the object or expressions used as input for the function.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@
 
 /go.work
 /go.work.sum
+.idea
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,4 @@
 ### Improvements
 
 ### Bug Fixes
+ - Convert expressions inside `jsonencode` calls without rewriting object keys to camelCase
--- a/pkg/convert/testdata/programs/jsonencode_preserves_casing/main.tf
+++ b/pkg/convert/testdata/programs/jsonencode_preserves_casing/main.tf
@@ -1,0 +1,12 @@
+# When converted, the object inside jsonencode should maintain the casing of the keys
+output "data" {
+  value = jsonencode({
+    "foo": "bar",
+    Content: "capitalized",
+    "Quoted": "quoted",
+    nested: [
+      { "Key": "value" },
+      { ANOTHER: "one" }
+    ]
+  })
+}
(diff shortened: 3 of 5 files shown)


scripts/generate_builtins.py
```diff
@@ -135,8 +137,7 @@ def trimext(file: str) -> str:
                 else:
                     example_code = [override]
             elif not example_code:
-                raise Exception(f"No examples found for {function_name}")
-
+                continue
```
Zaid-Ajaj, line 140: If we couldn't parse the code of the function, just skip. I don't think parsing the docs like this is reliable and is probably easier to just handwrite the functions we want to support 🤔

Frassle: > I don't think parsing the docs like this is reliable and is probably easier to just handwrite the functions we want to support 

Yes that's what the overrides list is for. But we need this error to tell us when something needs to be added to the overrides list. This should error.

scripts/generate_builtins.py
```diff
@@ -29,6 +29,8 @@ def trimext(file: str) -> str:
     "filesha512": "filesha512(\"hello.txt\")",
     "list": "list(1, 2, 3)",
     "map": "map(\"a\", \"b\", \"c\", \"d\")",
+    "jsonencode": "jsonencode({ \"foo\": \"bar\", Content: \"capitalized\", \"Quoted\": \"quoted\" })",
```
Frassle, line 32: Please don't abuse this automated broad test for testing specific things. This should have been written as a new specific test (jsonencode_preserve_casing or similar)

Zaid-Ajaj: That is a better idea, it also would have been easier than digging into why generate_builtins was failing

scripts/generate_builtins.py
Zaid-Ajaj: Alright, reverted my changes to this generator and added a test program for `jsonencode` ✅

Frassle (approved):

Issue #126, opened 2024-03-21
[docs] SQL DatabaseInstance docs indicate `depends_on` which does not appear in code blocks


kpitzen:
### What happened?

The docs here: https://www.pulumi.com/registry/packages/gcp/api-docs/sql/databaseinstance/ indicate that there should be an explicit `depends_on` in the resulting Pulumi code, but none exists.

### Example

> NOTE: For private IP instance setup, note that the gcp.sql.DatabaseInstance does not actually interpolate values from gcp.servicenetworking.Connection. You must explicitly add a depends_onreference as shown below.

```typescript
import * as pulumi from "@pulumi/pulumi";
import * as gcp from "@pulumi/gcp";
import * as random from "@pulumi/random";

const privateNetwork = new gcp.compute.Network("private_network", {name: "private-network"});
const privateIpAddress = new gcp.compute.GlobalAddress("private_ip_address", {
    name: "private-ip-address",
    purpose: "VPC_PEERING",
    addressType: "INTERNAL",
    prefixLength: 16,
    network: privateNetwork.id,
});
const privateVpcConnection = new gcp.servicenetworking.Connection("private_vpc_connection", {
    network: privateNetwork.id,
    service: "servicenetworking.googleapis.com",
    reservedPeeringRanges: [privateIpAddress.name],
});
const dbNameSuffix = new random.RandomId("db_name_suffix", {byteLength: 4});
const instance = new gcp.sql.DatabaseInstance("instance", {
    name: pulumi.interpolate`private-instance-${dbNameSuffix.hex}`,
    region: "us-central1",
    databaseVersion: "MYSQL_5_7",
    settings: {
        tier: "db-f1-micro",
        ipConfiguration: {
            ipv4Enabled: false,
            privateNetwork: privateNetwork.id,
            enablePrivatePathForGoogleCloudServices: true,
        },
    },
});
```

### Output of `pulumi about`

N/A

### Additional context

https://github.com/pulumi/pulumi-gcp/blob/d3120a011b56f81babc8a81f7546ad64c0125ad1/sdk/nodejs/sql/databaseInstance.ts#L43

This is the bridged docs file for typescript.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #127, opened 2024-03-26
Support depends_on option

Frassle:


Zaid-Ajaj (approved):

justinvp:
I assume this is for https://github.com/pulumi/pulumi-converter-terraform/issues/126?

PR #128, opened 2024-03-28
Emit warnings for non-supported lifecycle hooks `create_before_destroy` and `replace_triggered_by`

Zaid-Ajaj:
### Description

Partially addressing https://github.com/pulumi/pulumi-aws/issues/3482 by first emitting a warning from the converter when encountering these non-supported lifecycle hooks. Then in the bridge, we will do a pass over the diagnostics, detect these and turn them into errors so that the examples no longer will show up in the registry docs

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,6 +1,7 @@
 ### Improvements
 
 - Support the `depends_on` option
+- Emit warnings when encountering non-supported lifecycle hooks `create_before_destroy` and `replace_triggered_by` 
 
 ### Bug Fixes
  - Convert expressions inside `jsonencode` calls without rewriting object keys to camelCase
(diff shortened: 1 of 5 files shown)


Frassle (commented):
> [create_before_destroy](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#create_before_destroy) (bool) - By default, when Terraform must change a resource argument that cannot be updated in-place due to remote API limitations, Terraform will instead destroy the existing object and then create a new replacement object with the new configured arguments.

Huh this is actually the _opposite_ of Pulumi right? So shouldn't we 
A) Expose `deleteBeforeReplace` in PCL
B) Ignore `create_before_destroy=true` because that's the pulumi engines default
C) Turn every non-set or `create_before_destroy=false` to `deleteBeforeReplace=true`

The replace_triggered_by changes are fine.

Frassle (approved):
Eh fine, lets just leave them both as warning for now. We should probably re-visit this though.

PR #143, opened 2024-04-16
Cache the mapper

t0yv0:
Observing a lot of traffic on the local network between pulumi CLI and the converter plugin during AWS builds. Wanted to try this caching to see if this is about re-fetching the mappings repeatedly.

### DIFF
--- a/cmd/pulumi-converter-terraform/caching_mapper.go
+++ b/cmd/pulumi-converter-terraform/caching_mapper.go
@@ -1,0 +1,53 @@
+// Copyright 2016-2024, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package main
+
+import (
+	"context"
+	"sync"
+
+	"github.com/pulumi/pulumi/pkg/v3/codegen/convert"
+)
+
+func newCachingMapper(m convert.Mapper) convert.Mapper {
+	return &cachingMapper{
+		inner: m,
+		cache: make(map[string]map[string][]byte),
+	}
+}
+
+type cachingMapper struct {
+	inner convert.Mapper
+	mu    sync.Mutex
+	cache map[string]map[string][]byte
+}
+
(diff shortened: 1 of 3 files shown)


t0yv0:
For the AWS build: yes this does reduce network traffic. Can't say it speeds it up a lot; still CPU bound. Resident ram < 1GB, somewhere 600mb-800mb for me.

cmd/pulumi-converter-terraform/caching_mapper.go
```diff
@@ -0,0 +1,53 @@
...
+
+type cachingMapper struct {
+	inner convert.Mapper
+	mu    sync.Mutex
+	cache map[string]map[string][]byte
+}
+
+func (cm *cachingMapper) GetMapping(ctx context.Context, provider string, pulumiProvider string) ([]byte, error) {
+	cm.mu.Lock()
+	defer cm.mu.Unlock()
```
tgummerer, line 38: I was wondering if it would make sense to have a read-lock only here for cache hits, and then upgrade to a write-lock if we have a cache miss, but it looks like most of the code running this is serialized anyway, so it might just not be worth the complexity.

cmd/pulumi-converter-terraform/caching_mapper_test.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCachingMapper(t *testing.T) {
+	calls := 0
```
tgummerer, line 25:
```suggestion
	t.Parallel()

	calls := 0
```

Usually we have a linter for this, but I guess that's not set up in this repo.

cmd/pulumi-converter-terraform/caching_mapper_test.go
```diff
@@ -0,0 +1,65 @@
...
+		_, err := cm.GetMapping(context.Background(), "unknown", "unknown")
+		require.Error(t, err)
+		require.Equal(t, 1, calls)
+	})
+
+	t.Run("cache miss", func(t *testing.T) {
+		mappings, err := cm.GetMapping(context.Background(), "myprov", "myprov")
+		require.NoError(t, err)
+		require.Equal(t, "myprov-mappings", string(mappings))
+		require.Equal(t, 2, calls)
```
tgummerer, line 48: We should probably reset the `calls` variable at the beginning of each test case.  I think this is a little confusing, but also since you have a `t.Run()` here, someone could run this test as `go test -run 'TestCachingMapper/cache-miss`, and it would end up failing consistently, because the "error case" test would not have been run.

tgummerer (approved):
Please consider the nits inline.

t0yv0:
Thanks @tgummerer 

Actually since I don't get any speedup I think I'll just drop this - less code to maintain. What Fraser had in mind for precise lookups is the way to go.

PR #151, opened 2024-06-17
Add custom translation of `appautoscaling` ids

arnaubt:
These two resources have special import ids and were missing. Sources:
- `aws_appautoscaling_policy`: https://www.pulumi.com/registry/packages/aws/api-docs/appautoscaling/policy/#import
- `aws_appautoscaling_target`: https://www.pulumi.com/registry/packages/aws/api-docs/appautoscaling/target/#import

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,4 +1,5 @@
 ### Improvements
 
+- Support custom translation of appautoscaling ids
 
 ### Bug Fixes
--- a/pkg/convert/tf_state.go
+++ b/pkg/convert/tf_state.go
@@ -140,6 +140,40 @@ func TranslateState(info il.ProviderInfoSource, path string) (*plugin.ConvertSta
 						}
 
 						id = fmt.Sprintf("%s/%s", role, policy)
+					case "aws_appautoscaling_policy":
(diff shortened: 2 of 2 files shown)


tgummerer:
@arnaubt Thanks for your contribution!  I'm trying to get the acceptance tests for external contributions running properly here, sorry about the noise while I'm fixing that.  

In the meantime, I notice that there's a changelog entry missing for this change still.  Would you mind adding that to the `CHANGELOG_PENDING.md` file? The rest of this contribution looks good to me.

arnaubt:
@tgummerer Changelog entry added! Btw, do you have a rough idea when this could be released once merged?

tgummerer:
Thank you!  Once I manage to get the tests to run and merge this, I think we can create a new release.  Releases for this converter don't follow a schedule, but are rather done ad-hoc, whenever needed. I'll try to do it as soon as possible.  If everything goes well I'm hoping I can do a release today, but worst case by the end of the week we should have a new release.

tgummerer (approved):

tgummerer:
@arnaubt I just released v1.0.18 of pulumi-converter-terraform which includes this contribution. So you should be able to go ahead and use it now.  Thanks again!

arnaubt:
Thank you @tgummerer! That was a quick process.

PR #175, opened 2024-07-24
Add YAML field to translatedExample

Frassle:
Providers team would like to make use of the converted Pulumi.yaml files for provider config during bulk conversion. This adds a "yaml" field to the bulk conversion result for that.

Fixes https://github.com/pulumi/pulumi-converter-terraform/issues/176.

### DIFF
--- a/cmd/pulumi-converter-terraform/main.go
+++ b/cmd/pulumi-converter-terraform/main.go
@@ -59,6 +59,7 @@ func (*tfConverter) ConvertState(_ context.Context,
 
 type translatedExample struct {
 	PCL         string          `json:"pcl"`
+	YAML        string          `json:"yaml"`
 	Diagnostics hcl.Diagnostics `json:"diagnostics"`
 }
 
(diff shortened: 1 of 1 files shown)


cmd/pulumi-converter-terraform/main.go
```diff
@@ -107,15 +108,24 @@ func (*tfConverter) ConvertProgram(_ context.Context,
...
 			}
 
+			yml, err := afero.ReadFile(dst, "/Pulumi.yaml")
+			if err != nil && !os.IsNotExist(err) {
+				return translatedExample{}, fmt.Errorf("read example Pulumi.yaml %s from memory store: %w", name, err)
+			}
+
 			return translatedExample{
 				PCL:         string(pcl),
+				YAML:        string(yml),
```
lunaris, line 118: Minor paranoia: would we ever look at generating things like `Pulumi.deployment.yaml`s etc. or stack-specific YAMLs? If so then maybe we'd call this `pulumiYAML`? But if not then ignore me; looks good.

lunaris (approved):

cmd/pulumi-converter-terraform/main.go
Frassle: Probably not from the converter but happy to rename it just in case

Issue #177, opened 2024-07-24
Document `--convert-examples` option


guineveresaenger:
What it says on the box. We use this flag for bulk conversion HCL --> PCL in the bridge, but it's not obvious what this flag does or how to use it when encountered in the wild.

PR #182, opened 2024-08-01
Document --convert-examples

Frassle:
Fixes https://github.com/pulumi/pulumi-converter-terraform/issues/177.

As discussed with Will and Guinevere, add some comments to help explain --convert-examples and how it's used.


tgummerer (approved):

guineveresaenger (approved):
This makes sense; thank you for adding this!

PR #188, opened 2024-09-26
Fix using multiple modules with different version constraints

Frassle:
If two registry modules resolved to the same version but via different constraints we would error that the path was duplicated. This fixes that to allow both modules to share the same code.


pkg/convert/testdata/programs/registry_module_multiple/main.tf
```diff
@@ -0,0 +1,36 @@
+// This test checks that if we have the same module used twice, but it doesn't have _exactly_ the same version string
+// (and so a different key) that it still works correctly.
```
lunaris, line 2: TIL that TF supports `//` for single-line comments :exploding_head:

lunaris (approved):
LGTM

Issue #201, opened 2024-10-09
Converting terraform-azurerm-caf results in an error due to keyword overlap with "for"


brandonpollack23:
```
resource "mssqldb" "azurerm:index:mssqlServerExtendedAuditingPolicy" {
  options {
    dependsOn = [for]
    range     = { for key, value in myDatabase.mssqlDatabases : key => value if notImplemented("try(value.extended_auditing_policy,null)") != null }
  }
  logMonitoringEnabled               = notImplemented("try(each.value.extended_auditing_policy.log_monitoring_enabled,false)")
  serverId                           = mssqlServers[range.key].id
  storageEndpoint                    = mssqlAuditing[range.key].primaryBlobEndpoint
  storageAccountAccessKey            = mssqlAuditing[range.key].primaryAccessKey
  storageAccountAccessKeyIsSecondary = false
  retentionInDays                    = notImplemented("try(each.value.extended_auditing_policy.retention_in_days,null)")
}
```

is generated from 

```
resource "azurerm_mssql_server_extended_auditing_policy" "mssqldb" {
  depends_on = [azurerm_role_assignment.for]
  for_each = {
    for key, value in local.database.mssql_databases : key => value
    if try(value.extended_auditing_policy, null) != null
  }


  log_monitoring_enabled                  = try(each.value.extended_auditing_policy.log_monitoring_enabled, false)
  server_id                               = module.mssql_servers[each.key].id
  storage_endpoint                        = data.azurerm_storage_account.mssql_auditing[each.key].primary_blob_endpoint
  storage_account_access_key              = data.azurerm_storage_account.mssql_auditing[each.key].primary_access_key
  storage_account_access_key_is_secondary = false
  retention_in_days                       = try(each.value.extended_auditing_policy.retention_in_days, null)
}
```

The for is defined in roles.tf:
```
resource "azurerm_role_assignment" "for" {
  for_each = {
    for key, value in try(local.roles_to_process, {}) : key => value
    if contains(keys(local.services_roles), value.scope_resource_key)
  }

  principal_id         = each.value.object_id_resource_type == "object_ids" ? each.value.object_id_key_resource : each.value.object_id_lz_key == null ? local.services_roles[each.value.object_id_resource_type][var.current_landingzone_key][each.value.object_id_key_resource].rbac_id : local.services_roles[each.value.object_id_resource_type][each.value.object_id_lz_key][each.value.object_id_key_resource].rbac_id
[...]

PR #207, opened 2024-10-11
Fix keyword overlap in terraform names (specifically for)

brandonpollack23:
Right now it detects types and names and appends the kind and "pcl" to
help distinguish the generated types and minimize chance for overlap.

It also saves these renames to the convert state and uses them to rename
references.  Type renames are saved but not currently used.

#201

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,7 +1,7 @@
 ### Improvements
 
-- Added references to issues for missing functions in output.
-- Added code generation rename workarounds for pcl keywords.
+- Add references to issues for missing functions in output
+- Add code generation rename workarounds for pcl keywords
 
 ### Bug Fixes
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,6 +1,7 @@
 ### Improvements
 
 - Added references to issues for missing functions in output.
+- Added code generation rename workarounds for pcl keywords.
 
 ### Bug Fixes
 
--- a/pkg/convert/testdata/programs/name_overlap/pcl/main.pp
+++ b/pkg/convert/testdata/programs/name_overlap/pcl/main.pp
@@ -1,0 +1,13 @@
+resource "forResourcePcl" "simple:index:resource" {
+  __logicalName = "for_resource_pcl"
+  inputOne      = "hello"
+  inputTwo      = true
+}
+
+resource "dependsOnFor" "simple:index:resource" {
+  options {
+    dependsOn = [forResourcePcl]
+  }
+  inputOne = forResourcePcl.result
+  inputTwo = false
+}
--- a/pkg/convert/tf.go
+++ b/pkg/convert/tf.go
@@ -765,13 +765,21 @@ type convertState struct {
 
 	// Determines whether converting objects should rewrite keys to camelCase or keep it as is
 	rewriteObjectKeys bool
+
+	// Names that were changed to avoid overlaps with pcl keywords.
+	renames     pclOverlapRenames
+	typeRenames pclOverlapRenames
 }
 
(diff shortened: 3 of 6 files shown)


CHANGELOG_PENDING.md
```diff
@@ -1,6 +1,7 @@
 ### Improvements
 
 - Added references to issues for missing functions in output.
+- Added code generation rename workarounds for pcl keywords.
```
tgummerer, line 4:
```suggestion
- Add references to issues for missing functions in output
- Add code generation rename workarounds for pcl keywords
```

Both of the entries here don't conform to the chanagelog style we adopted in pulumi/pulumi.  We should probably be consistent with that here as well.  See https://github.com/pulumi/pulumi/blob/master/CONTRIBUTING.md#changelog-messages.

pkg/convert/tf.go
```diff
@@ -787,6 +795,58 @@ func (s *convertState) sourceCode(rng hcl.Range) string {
...
+
+	if isPclKeyword(name) {
+		newName = fmt.Sprintf("%s_%s_pcl", name, kind)
+		s.renames.nameToRename[name] = newName
+
+		s.appendDiagnostic(&hcl.Diagnostic{
+			Subject:  hclRange,
+			Severity: hcl.DiagWarning,
+			Summary:  kind + " renamed to prevent keyword overlap",
+			Detail:   fmt.Sprintf("Renaming %s %s to of type %s to %s to prevent overlap",kind,name, *hclType, newName),
```
Zaid-Ajaj, line 827:
```diff
- Renaming %s %s to
+ Renaming %s %s
```

Zaid-Ajaj (approved):
Looks great! just a small typo in the warning message

pkg/convert/testdata/programs/name_overlap/pcl/main.pp
```diff
@@ -0,0 +1,13 @@
+resource "forResourcePcl" "simple:index:resource" {
+  __logicalName = "for_resource_pcl"
```
justinvp, line 2: Does the logical name need to be renamed? Couldn't it remain the same?

pkg/convert/tf.go
```diff
@@ -787,6 +795,58 @@ func (s *convertState) sourceCode(rng hcl.Range) string {
...
+		s.appendDiagnostic(&hcl.Diagnostic{
+			Subject:  hclRange,
+			Severity: hcl.DiagWarning,
+			Summary:  kind + " type renamed to prevent keyword overlap",
+			Detail:   fmt.Sprintf("Renaming %s type %s to %s to prevent overlap", kind, *hclType, *newType),
+		})
+	}
+
+	if isPclKeyword(name) {
+		newName = fmt.Sprintf("%s_%s_pcl", name, kind)
```
justinvp, line 820: PCL isn't really something we've publicly documented. It's really more of an internal implementation detail. I'm wondering if we should use something other than "pcl" for the suffix? Would a single underscore work or does it need to be alpha numeric?

pkg/convert/testdata/programs/name_overlap/pcl/main.pp
brandonpollack23: I think it might be better just to keep it consistent (and they are kind of generated from the same field so it'd take more work to make them diverge) which seems like work for confusion.

pkg/convert/tf.go
brandonpollack23: I'll give it a shot

pkg/convert/testdata/programs/name_overlap/pcl/main.pp
Zaid-Ajaj: @brandonpollack23 it might make sense to keep the logical name as `for` if people are converting their code, then importing the state from tf state then the references will match

PR #211, opened 2024-10-15
Fix keyword overlap in terraform names (specifically for) (#207)

brandonpollack23:
Right now it detects types and names and appends the kind and "pcl" to
help distinguish the generated types and minimize chance for overlap.

It also saves these renames to the convert state and uses them to rename
references.  Type renames are saved but not currently used.

#201

---------

Co-authored-by: Thomas Gummerer <t.gummerer@gmail.com>


tgummerer:
I'm confused by this PR, wasn't this what was merged in #207?

Issue #181, opened 2024-07-30
Converter should handle variables that have similar names to TF resources and functions


pierskarsenbarg:
### What happened?

Trying to convert this TF program: https://github.com/aws-ia/terraform-aws-vpc

Program references a third-party module: https://github.com/aws-ia/terraform-aws-vpc/blob/8adf507bbddb5d959fc63cb894f40afc9d9f4405/modules/calculate_subnets/main.tf#L26

Module uses local variable called `new_bits`: https://github.com/drewmullen/terraform-cidr-subnets/blob/eaa783efe131fb740a30f5ad8b4382187461f345/main.tf#L3

The converter sees this as being related to `NewBits` (https://github.com/pulumi/pulumi-std/blob/79ed7e4d84cff8a09195089930a7edd20bdcaaf7/std/cidrsubnet.go#L31) and tries to rename it which causes the error

Error message:

```
================================================================================
The Pulumi CLI encountered a code generation error. This is a bug!
We would appreciate a report: https://github.com/pulumi/pulumi/issues/
Please provide all of the below text in your report.
================================================================================
Pulumi Version:   v3.127.0
[...]

PR #213, opened 2024-10-18
Handle snake case in string literal map keys curing conversion.

brandonpollack23:
Fixes: #181

### DIFF
--- a/pkg/convert/testdata/programs/expressions/pcl/main.pp
+++ b/pkg/convert/testdata/programs/expressions/pcl/main.pp
@@ -66,9 +66,9 @@ output "complexObjectOut" {
   value = {
     aTuple = ["a", "b", "c"]
     anObject = {
-      literalKey                = 1
-      anotherLiteralKey         = 2
-      "yet_another_literal_key" = aValue
+      literalKey             = 1
+      anotherLiteralKey      = 2
+      "yetAnotherLiteralKey" = aValue
 
       // This only translates correctly in the new converter.
       (aKey) = 4
--- a/pkg/convert/tf.go
+++ b/pkg/convert/tf.go
@@ -1107,6 +1107,18 @@ func convertObjectConsKeyExpr(state *convertState, inBlock bool,
 	scopes *scopes, fullyQualifiedPath string, expr *hclsyntax.ObjectConsKeyExpr,
 ) hclwrite.Tokens {
 	// Seems we can just ignore ForceNonLiteral here
+	switch exprWrapped := expr.Wrapped.(type) {
+	case *hclsyntax.TemplateExpr:
+		// For some reason these are template expressions, not just literals,
+		// extracing the static string.
+		name, _ := matchStaticString(exprWrapped)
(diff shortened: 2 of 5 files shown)


pkg/convert/testdata/programs/expressions/pcl/main.pp
```diff
@@ -66,9 +66,9 @@
   value = {
     aTuple = ["a", "b", "c"]
     anObject = {
-      literalKey                = 1
-      anotherLiteralKey         = 2
-      "yet_another_literal_key" = aValue
+      literalKey             = 1
+      anotherLiteralKey      = 2
+      "yetAnotherLiteralKey" = aValue
```
Zaid-Ajaj, line 71: This change doesn't look right, a literal key should probably stay as is 🤔 we only rewrite object keys such as these when we know we can do that: for example terraform keys of snake_case matching camelCase pulumi fields. 

We would need to apply this rewrite whenever the program references it somewhere down the line:
```tf
locals {
  obj = {
    "snake_case_key_becomes_camel" = "abc123",
  }
}

output "referenced" {
  // is the referenced key rewritten here too? maybe we shouldn't actually rewrite these
  value = local.obj["snake_case_key_becomes_camel"]
}
```

pkg/convert/tf.go
```diff
@@ -1107,6 +1107,18 @@ func convertObjectConsKeyExpr(state *convertState, inBlock bool,
 	scopes *scopes, fullyQualifiedPath string, expr *hclsyntax.ObjectConsKeyExpr,
 ) hclwrite.Tokens {
 	// Seems we can just ignore ForceNonLiteral here
+	switch exprWrapped := expr.Wrapped.(type) {
+	case *hclsyntax.TemplateExpr:
+		// For some reason these are template expressions, not just literals,
```
Zaid-Ajaj, line 1112: > For some reason these are template expressions, not just literals,

I believe a template expression is a string that can have literal _parts_ (think interpolated strings) but if it is a single string then it is a template expression with a single part being the literal value

pkg/convert/testdata/programs/expressions/pcl/main.pp
brandonpollack23: This is exactly the concern I had when writing it (I keep forgetting to mark things as draft when it is just an RFC).

The solution then would be to detect whenever we have a literal in a map and save it in a list of things NOT to mangle and then be sure to check it.  WDYT?

Zaid-Ajaj: I don't know about keeping track of things, instead _knowing_ when we should actually mangle or not and tend not to mangle when possible. Only mangle reference names, rewrite to match pulumi fields, otherwise keep it as is 🤔 ?

brandonpollack23: yeah I think that will work

brandonpollack23:
replaced with attempt 2: #217

PR #217, opened 2024-10-30
Change object keys to not be mangled when converting from terraform

brandonpollack23:
- **Handle snake case in string literal map keys curing conversion.**
- **fix bad quoting**
- **initial accept, things are broken (like dynamic blocks)**


justinvp:
How will this change impact actual programs (TypeScript, Go, C#, etc.)?

brandonpollack23:
> How will this change impact actual programs (TypeScript, Go, C#, etc.)?

Good question, I tested it locally and it does keep the snake case names in other languages.

The thing is, this is sort of up to the converters for each language from PCL to know "a local should be camel" or snake etc.

Right now we do the wrong thing in languages like python we do it wrong since PEP8 wants snake (unless the converter already does that).

Actually because of these bugs these programs won't even get past pcl...

pkg/convert/testdata/programs/dynamic_max_items_one/pcl/diagnostics.json
```diff
@@ -0,0 +1,3 @@
+[
+  "warning:dynamic_max_items_one/main.tf:3,20-26:dynamic_max_items_one/main.tf:3,20-26:unexpected expression in loop collection:expected a collection expression *hclsyntax.TupleConsExpr in the loop traversal, but found an unexpected expression"
+]
```
Zaid-Ajaj, line 3: Do we know why suddenly these diagnostics are showing up here?

Zaid-Ajaj:
> How will this change impact actual programs (TypeScript, Go, C#, etc.)?

> Good question, I tested it locally and it does keep the snake case names in other languages.

> The thing is, this is sort of up to the converters for each language from PCL to know "a local should be camel" or snake etc.

> Right now we do the wrong thing in languages like python we do it wrong since PEP8 wants snake (unless the converter already does that).

@brandonpollack23 We probably need the language program generators to fix how they handle local variables renames before merging this if we don't want large diffs on provider upgrades. i.e. C#/Go turning `input_value = ...` into `var inputValue = ...`

justinvp:
> @brandonpollack23 We probably need the language program generators to fix how they handle local variables renames before merging this if we don't want large diffs on provider upgrades. i.e. C#/Go turning `input_value = ...` into `var inputValue = ...`

This is my main concern. While the change will allow more things to be converted, existing converting programs are going to regress. Converted TF examples that used to look idiomatic will now look unnatural (in languages that don't typically use snake_case). I'm not sure fixing each language program generator is the right approach.

brandonpollack23:
Yeah I'm wondering if we can fix this on the gen side, like in line: 579 of pulumi/pulumi/pkg/codegen/go/gen_program_expressions.go just detect if it is snake and make it camel (for each language).  This includes ObjectConsExpr and I think there might be some others but it shouldn't be too difficult.

pkg/convert/testdata/programs/object_keys_renamed/main.tf
```diff
@@ -0,0 +1,14 @@
+locals {
```
Frassle, line 1: I think this test is valid. We should be able to raise a PR with just this new test, and without any other changes to the other generated files.

brandonpollack23:
So we discussed in a meeting, we decided it is better to not mangle Object keys, but we should mangle locals always and then in another step we should use the type information we have on assignment to assign the correct values per key in a resource (generate the right stuff)

PR #218, opened 2024-11-01
Handles missing expression type TemplateJoinExpr in convertExpression.

brandonpollack23:
Without handling this we panic, so instead lets not do that and handle
this by unwrapping the inner ForExpr.

Related #212

### DIFF
--- a/pkg/convert/tf.go
+++ b/pkg/convert/tf.go
@@ -1141,10 +1141,68 @@ func convertTemplateWrapExpr(state *convertState,
 	return tokens
 }
 
+func convertTemplateJoinExpr(state *convertState, inBlock bool, scopes *scopes,
+	fullyQualifiedPath string, expr *hclsyntax.TemplateJoinExpr,
+) hclwrite.Tokens {
+	forExpr := expr.Tuple.(*hclsyntax.ForExpr)
+	if forExpr == nil {
+		contract.Failf("Template Join Expression must contain for expression, but contains %T", expr.Tuple)
+	}
+
+	// The collection doesn't yet have access to the key/value scopes
+	collTokens := convertExpression(state, false, scopes, "", forExpr.CollExpr)
+
+	// TODO: We should ensure key and value vars are unique
+	locals := map[string]string{
+		forExpr.ValVar: camelCaseName(forExpr.ValVar),
+	}
+	if forExpr.KeyVar != "" {
+		locals[forExpr.KeyVar] = camelCaseName(forExpr.KeyVar)
+	}
+	scopes.push(locals)
+
+	valueTokens := convertExpression(state, false, scopes, "", forExpr.ValExpr)
+	valueTokens = valueTokens[1 : len(valueTokens)-1] // strip the outer quote tokens
+
+	scopes.pop()
+
+	var tokens hclwrite.Tokens
+	tokens = append(tokens, makeToken(hclsyntax.TokenTemplateControl, "%{"))
+
+	// Write the for template control section.
+	tokens = append(tokens, makeToken(hclsyntax.TokenIdent, "for"))
+	if locals[forExpr.KeyVar] != "" {
+		tokens = append(tokens, makeToken(hclsyntax.TokenIdent, locals[forExpr.KeyVar]))
+		tokens = append(tokens, makeToken(hclsyntax.TokenComma, ","))
+	}
+	tokens = append(tokens, makeToken(hclsyntax.TokenIdent, locals[forExpr.ValVar]))
+	tokens = append(tokens, makeToken(hclsyntax.TokenIdent, "in"))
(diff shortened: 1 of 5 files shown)


tests/example_test.go
```diff
@@ -129,7 +129,7 @@ func TestExample(t *testing.T) {
 		{
 			example: "https://github.com/terraform-aws-modules/terraform-aws-vpc",
 			// TODO[pulumi/pulumi#13743]: unknown property 'domain' among [...]
-			skip: allLanguages,
+			skip: newStringSet(csharp, golang, python, typescript),
```
tgummerer, line 132: This is still `allLanguages`.  Why did we change this, `allLanguages` seems more expressive to me?

But really what alerted me to look at this change is that we should probably add a test for the behaviour of  `TemplateJoinExpr`.

brandonpollack23:
After some offline discussion, I want to color in the problem and why the current state of the change won't fix it:

This operator is caused by these types of expressions:
```hcl
%{for metric in var.cilium_hubble_metrics_enabled~}
      - "${metric}"
%{endfor~}
```

Which basically join all these templated strings together, I think I can fix it simililary to other template exprs...

brandonpollack23:
I've pushed an update that should handle this but there is still a missing piece.  We assume all strings are regular "" delimited, but dont seem to handle EOF strings, causing our output to be unescaped and ruin the parsing of PCL.  Investigating this separately in #219

.vscode/launch.json
```diff
@@ -4,12 +4,20 @@
...
-            "mode": "local",
-            "processId": 31790
-        }
+      {
+        "name": "Launch test function",
+        "type": "go",
+        "request": "launch",
+        "mode": "test",
+        "program": "${workspaceFolder}/pkg/convert",
+        "args": ["-test.run", "TestTranslate/template_join_expr"]
```
lunaris, line 13: Accidental add?

pkg/convert/tf.go
```diff
@@ -1141,10 +1141,68 @@ func convertTemplateWrapExpr(state *convertState,
 	return tokens
 }
 
+func convertTemplateJoinExpr(state *convertState, inBlock bool, scopes *scopes,
+	fullyQualifiedPath string, expr *hclsyntax.TemplateJoinExpr,
+) hclwrite.Tokens {
+	forExpr := expr.Tuple.(*hclsyntax.ForExpr)
+	if forExpr == nil {
+		contract.Failf("Template Join Expression must contain for expression, but contains %T", expr.Tuple)
+	}
```
lunaris, line 1150: Are all TemplateJoinExprs for-expressions?

lunaris (commented):
This looks similar to #219, so maybe the comments I've left there apply here? Or the branches are stacked? Apologies if I'm being dumb here :see_no_evil:

brandonpollack23:
> This looks similar to #219, so maybe the comments I've left there apply here? Or the branches are stacked? Apologies if I'm being dumb here 🙈

They're stacked :)

.vscode/launch.json
brandonpollack23: I thought it might be ok but yeah it is kind of specific

pkg/convert/tf.go
brandonpollack23: They are :)

tgummerer:
> They're stacked :)

You can set the base branch for stacked PRs, to the branch they are stacked on, so earlier commits don't show up to avoid the confusion.   Also it's helpful to point out which PR something is stacked on.

brandonpollack23:
Set the base to make it easier to see stacks, thanks for letting me know (didn't know that was possible!)

tests/example_test.go
```diff
@@ -121,6 +121,11 @@ func TestExample(t *testing.T) {
 		commit  string
 		skip    stringSet
 	}{
+		{
+			example: "https://github.com/kube-hetzner/terraform-hcloud-kube-hetzner",
+			// TODO[pulumi/pulumi-converter-terraform#212]: Hetzner cloud has various conversion issues.
+			skip: allLanguages,
+		},
```
lunaris, line 128: Why are we adding it if it's always skipped? Would it be better to add it when we fix it?

brandonpollack23:
Right now we have an active effort to remove these skips, adding it here
will ensure we don't forget to make these pass as well, but I guess that
can be tracked by the bug (if it doesn't get lost in the ocean)

tests/example_test.go
brandonpollack23: Right now we're actively trying to unblock these, I just want to make sure these cases aren't forgotten and tracked in the code as well.  I can remove it and just use the bug but don't want to accidentally leave this user behind

Frassle: Yeh most of the others below are all skipped as well, it's fine to add to track it

PR #220, opened 2024-11-04
Handle heredoc string delimiters

### LINKED ISSUE #219: <<EOT EOT delimited strings (eg large multiline templates) not handled correctly and therefore unescaped in PCL
Related to #212 

One thing was fixed in not handling for templated strings, but now we must also consider EOT delimited raw strings.

brandonpollack23:
Maintain HEREDOC (<<EOT style) strings when converting.

Previously our template and string literals only handled strings
delimited by quotes, this will inspect the source and maintain the same
string delim to handle non escaped and multiline strings.

Fixes #219

### DIFF
--- a/pkg/convert/testdata/programs/expressions/pcl/main.pp
+++ b/pkg/convert/testdata/programs/expressions/pcl/main.pp
@@ -88,7 +88,17 @@ output "quotedTemplate" {
 }
 
 output "heredoc" {
-  value = "This is also a template.\nSo we can output the key again ${aKey}\n"
+  value = <<END
+This is also a template.
+So we can output the key again ${aKey}
+END
+}
+
+output "heredocWithHyphen" {
+  value = <<-END
+This is also a template.
+So we can output the key again ${aKey}
+END
 }
 
 output "forTuple" {
--- a/pkg/convert/testdata/programs/heredoc_template_expr/main.tf
+++ b/pkg/convert/testdata/programs/heredoc_template_expr/main.tf
@@ -1,0 +1,8 @@
+locals {
+  a_list = ["a", "b", "c"]
+  join_template_expr = <<EOT
+%{for v in local.a_list~}
+${v}
+%{endfor~}
+EOT
+}
--- a/pkg/convert/testdata/programs/heredoc_template_expr/pcl/main.pp
+++ b/pkg/convert/testdata/programs/heredoc_template_expr/pcl/main.pp
@@ -1,0 +1,4 @@
+aList            = ["a", "b", "c"]
+joinTemplateExpr = <<EOT
+%{for v in aList~}${v}\n%{endfor~}
+EOT
--- a/pkg/convert/tf.go
+++ b/pkg/convert/tf.go
@@ -22,6 +22,7 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 
@@ -1141,18 +1142,20 @@ func convertTemplateWrapExpr(state *convertState,
 	return tokens
 }
 
+// convertTemplateJoinExpr writes "template joins" back out as they are.  These
+// are for expressions in the HCL template syntax like so:
+// "%{for foo in local.bars~}foo%{endfor~}
 func convertTemplateJoinExpr(state *convertState, inBlock bool, scopes *scopes,
 	fullyQualifiedPath string, expr *hclsyntax.TemplateJoinExpr,
 ) hclwrite.Tokens {
-	forExpr := expr.Tuple.(*hclsyntax.ForExpr)
-	if forExpr == nil {
+	forExpr, ok := expr.Tuple.(*hclsyntax.ForExpr)
+	if !ok {
 		contract.Failf("Template Join Expression must contain for expression, but contains %T", expr.Tuple)
 	}
 
 	// The collection doesn't yet have access to the key/value scopes
 	collTokens := convertExpression(state, false, scopes, "", forExpr.CollExpr)
 
-	// TODO: We should ensure key and value vars are unique
 	locals := map[string]string{
 		forExpr.ValVar: camelCaseName(forExpr.ValVar),
 	}
@@ -1167,9 +1170,9 @@ func convertTemplateJoinExpr(state *convertState, inBlock bool, scopes *scopes,
 	scopes.pop()
 
 	var tokens hclwrite.Tokens
-	tokens = append(tokens, makeToken(hclsyntax.TokenTemplateControl, "%{"))
 
-	// Write the for template control section.
+	// Write the opening of the for template
+	tokens = append(tokens, makeToken(hclsyntax.TokenTemplateControl, "%{"))
 	tokens = append(tokens, makeToken(hclsyntax.TokenIdent, "for"))
 	if locals[forExpr.KeyVar] != "" {
 		tokens = append(tokens, makeToken(hclsyntax.TokenIdent, locals[forExpr.KeyVar]))
@@ -1178,19 +1181,12 @@ func convertTemplateJoinExpr(state *convertState, inBlock bool, scopes *scopes,
 	tokens = append(tokens, makeToken(hclsyntax.TokenIdent, locals[forExpr.ValVar]))
 	tokens = append(tokens, makeToken(hclsyntax.TokenIdent, "in"))
 	tokens = append(tokens, collTokens...)
-
 	tokens = append(tokens, makeToken(hclsyntax.TokenTemplateControl, "~}"))
 
-	// There is no key expr in template join expressions.
-
-	// Write the value part
+	// Write the inside of the for template
 	tokens = append(tokens, valueTokens...)
 
-	// There is no ellipsis in template join expressions.
-
-	// There is no conditional part in template join expressions.
-
-	// Write the endfor control section.
+	// Write the %{endfor~}
 	tokens = append(tokens, makeToken(hclsyntax.TokenTemplateControl, "%{"))
 	tokens = append(tokens, makeToken(hclsyntax.TokenIdent, "endfor"))
 	tokens = append(tokens, makeToken(hclsyntax.TokenTemplateControl, "~}"))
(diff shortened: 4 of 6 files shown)


.vscode/launch.json
```diff
@@ -4,12 +4,20 @@
...
-            "mode": "local",
-            "processId": 31790
-        }
+      {
+        "name": "Launch test function",
+        "type": "go",
+        "request": "launch",
+        "mode": "test",
+        "program": "${workspaceFolder}/pkg/convert",
+        "args": ["-test.run", "TestTranslate/heredoc_template_expr"]
```
lunaris, line 13: Accidental inclusion?

pkg/convert/testdata/programs/heredoc_template_expr/pcl/main.pp
```diff
@@ -0,0 +1,4 @@
+aList            = ["a", "b", "c"]
+joinTemplateExpr = <<EOT
+%{for v in aList~}${v}\n%{endfor~}
+EOT
```
lunaris, line 4: Just sanity checking, sorry if dumb question -- do we want it realised this way, with the `\n` in the literal, or should we actually write out line breaks?

pkg/convert/tf.go
```diff
@@ -1141,14 +1141,77 @@ func convertTemplateWrapExpr(state *convertState,
...
+) hclwrite.Tokens {
+	forExpr := expr.Tuple.(*hclsyntax.ForExpr)
+	if forExpr == nil {
+		contract.Failf("Template Join Expression must contain for expression, but contains %T", expr.Tuple)
+	}
+
+	// The collection doesn't yet have access to the key/value scopes
+	collTokens := convertExpression(state, false, scopes, "", forExpr.CollExpr)
+
+	// TODO: We should ensure key and value vars are unique
```
lunaris, line 1155: Is it straightforward to do that here or do we need a follow-up issue?

pkg/convert/tf.go
```diff
@@ -1141,14 +1141,77 @@ func convertTemplateWrapExpr(state *convertState,
 	return tokens
 }
 
+func convertTemplateJoinExpr(state *convertState, inBlock bool, scopes *scopes,
+	fullyQualifiedPath string, expr *hclsyntax.TemplateJoinExpr,
+) hclwrite.Tokens {
+	forExpr := expr.Tuple.(*hclsyntax.ForExpr)
+	if forExpr == nil {
```
lunaris, line 1148: Won't this trigger a panic if the assertion fails? I feel like we need the two-value form, `forExpr, ok := ...`? I might be missing something though.

pkg/convert/tf.go
```diff
@@ -1141,14 +1141,77 @@ func convertTemplateWrapExpr(state *convertState,
...
+	// There is no key expr in template join expressions.
+
+	// Write the value part
+	tokens = append(tokens, valueTokens...)
+
+	// There is no ellipsis in template join expressions.
+
+	// There is no conditional part in template join expressions.
+
+	// Write the endfor control section.
```
lunaris, line 1193: These comments are (at least for me) generating more questions than they are answers. I wonder if it's worth instead adding a chunky-ish doc comment to the function that talks about what comes in, and what comes out, potentially with a sort of "code to code" example? So TF in that corresponds to a TemplateJoinExpr and tokens out that correspond to some PCL?

pkg/convert/tf.go
```diff
@@ -1141,14 +1141,77 @@ func convertTemplateWrapExpr(state *convertState,
...
+
+	return tokens
+}
+
 func convertTemplateExpr(state *convertState,
 	scopes *scopes, fullyQualifiedPath string, expr *hclsyntax.TemplateExpr,
 ) hclwrite.Tokens {
 	tokens := []*hclwrite.Token{}
-	tokens = append(tokens, makeToken(hclsyntax.TokenOQuote, "\""))
+	// TODO(#219) handle <<EOF EOF string expressions.
```
lunaris, line 1205: Isn't that what we're implementing?

pkg/convert/tf.go
```diff
@@ -1163,16 +1226,42 @@ func convertTemplateExpr(state *convertState,
...
+	start := r.Start.Byte
+	end := start + len("<<EOT")
+	end = min(end, len(file))
+	token := file[start:end]
+
+	if token[0] == '"' {
+		return false
+	}
+
+	if string(token) == "<<EOT" {
```
lunaris, line 1258: The delimiter in a heredoc can be anything, right? Not sure if TF has some limitations here, but I feel like you could do e.g.

```
<<FOO
...
FOO
```

It's just at least two `<`s (possibly `<<-` is also supported for slurping)

pkg/convert/testdata/programs/heredoc_template_expr/main.tf
```diff
@@ -0,0 +1,8 @@
+locals {
+  a_list = ["a", "b", "c"]
+  join_template_expr = <<EOT
+%{for v in local.a_list~}
+${v}
+%{endfor~}
+EOT
+}
```
lunaris, line 8: Worth adding some more test cases with different heredoc delimiters, heredocs that are "just" simple strings (no control blocks), etc.?

pkg/convert/tf.go
brandonpollack23: I think checking nil works, but might as well use hte ok feature

pkg/convert/tf.go
brandonpollack23: It actually leaves it the same, I'll just add the syntax up there that's probably going to help

pkg/convert/tf.go
brandonpollack23: yessir, just didnt remove it

pkg/convert/tf.go
brandonpollack23: Indeed, https://developer.hashicorp.com/terraform/language/expressions/strings#heredoc-strings

brandonpollack23: I added a fancy regex and return delimiters now

pkg/convert/testdata/programs/heredoc_template_expr/main.tf
brandonpollack23: This is included in the "expressions" already

pkg/convert/tf.go
```diff
@@ -1231,10 +1231,32 @@ func convertTemplateExpr(state *convertState,
...
+	} else {
+		tokens = append(tokens, makeToken(hclsyntax.TokenCQuote, "\""))
+	}
 	return tokens
 }
 
+// detectHeredocDelim detects if the range contains a heredoc token and returns
+// the open and close delimiters.
+func detectHeredocDelim(state *convertState, r hcl.Range) (string, string, bool) {
+	hereDocRegex := regexp.MustCompile(`^<<-?([[:alpha:]]+)\n`)
```
lunaris, line 1245: I think we might need to do better than this. If there's a hyphen, we need to trim leading whitespace from each line (assuming we aren't doing this elsewhere). Also, since `hereDocRegex` is a constant, it might be worth hoisting out (also for consistency I guess it should be ` heredocRegex`?).

Notes on `<<-`:

> If a heredoc template is introduced with the <<- symbol, any literal string at the start of each line is analyzed to find the minimum number of leading spaces, and then that number of prefix spaces is removed from all line-leading literal strings. The final closing marker may also have an arbitrary number of spaces preceding it on its line.

(From https://github.com/hashicorp/hcl/blob/main/hclsyntax/spec.md)

pkg/convert/tf.go
```diff
@@ -1141,18 +1142,20 @@ func convertTemplateWrapExpr(state *convertState,
...
+// convertTemplateJoinExpr writes "template joins" back out as they are.  These
+// are for expressions in the HCL template syntax like so:
+// "%{for foo in local.bars~}foo%{endfor~}
 func convertTemplateJoinExpr(state *convertState, inBlock bool, scopes *scopes,
 	fullyQualifiedPath string, expr *hclsyntax.TemplateJoinExpr,
 ) hclwrite.Tokens {
-	forExpr := expr.Tuple.(*hclsyntax.ForExpr)
-	if forExpr == nil {
+	forExpr, ok := expr.Tuple.(*hclsyntax.ForExpr)
+	if !ok {
```
lunaris, line 1152: I'm still not convinced here -- couldn't a template also be e.g. an `IfExpr`? https://developer.hashicorp.com/terraform/language/expressions/strings#string-templates 

Or do we handle those elsewhere/am I missing the point sorry

pkg/convert/tf.go
brandonpollack23: I don't think we need to trim the whitespace, the HCL parser will do that when the PCL is reparsed, won't it?  We are just regenerating the PCL compliant HCL, which is parsed again later with HCL utilities anyways.  No need for us to remove the leading whitespace.

pkg/convert/tf.go
brandonpollack23: This is a different kind of expression (though I'm not sure which).  I added that case to the template_join test though

pkg/convert/testdata/programs/heredoc_template_expr/pcl/main.pp
brandonpollack23: I believe we want it realized this way because then it will be reparsed identically

pkg/convert/tf.go
Frassle: Aye, we just need to emit it in exactly the same way

pkg/convert/testdata/programs/expressions/pcl/main.pp
```diff
@@ -88,7 +88,15 @@
 }
 
 output "heredoc" {
-  value = "This is also a template.\nSo we can output the key again ${aKey}\n"
+  value = <<END
+This is also a template.\nSo we can output the key again ${aKey}\n
```
Frassle, line 92: Won't this keep \n in the final result now because it's now wrapped in a HEREDOC?

brandonpollack23: great catch, didn't even notice.  Fixed now

pkg/convert/testdata/programs/expressions/pcl/main.pp
```diff
@@ -88,7 +88,21 @@
 }
 
 output "heredoc" {
-  value = "This is also a template.\nSo we can output the key again ${aKey}\n"
+  value = <<END
+This is also a template.
+So we can output the key again ${aKey}
+
```
Frassle, line 94: Lots of extra newlines here?

brandonpollack23: Didn't push...here we go

pkg/convert/testdata/programs/expressions/pcl/main.pp
```diff
@@ -88,7 +88,21 @@
...
+
```
Frassle, line 95: Extra whitespace still here

pkg/convert/testdata/programs/expressions/pcl/main.pp
```diff
@@ -88,7 +88,21 @@
...
+END
+}
+
+output "heredocWithHyphen" {
+  value = <<-END
+This is also a template.
+So we can output the key again ${aKey}
+
```
Frassle, line 103: And here

PR #229, opened 2024-11-26
Fix blocks of maps of resources

Frassle:
It was pointed out that in the TF shim schema an attribute that's typed
as a map but with a resource element should actually be treated as a
single object:
https://github.com/pulumi/pulumi-terraform-bridge/blob/f6e5260a88eab21043912e374bbafd5f42fd7f44/pkg/tfshim/shim.go#L149

This adds a test for schemas and programs of this shape.

### DIFF
--- a/pkg/convert/tf_scopes.go
+++ b/pkg/convert/tf_scopes.go
@@ -375,10 +375,18 @@ func (s *scopes) maxItemsOne(fullyQualifiedPath string) bool {
 		return *schemaInfo.MaxItemsOne
 	}
 
-	// If we have a shim schema use it's MaxItems
+	// If we have a shim schema use it's MaxItems and Type
 	sch := info.Schema
 	if sch != nil {
-		return sch.MaxItems() == 1
+		// If this is a list or set, check if it has a maxItems of 1.
+		if sch.Type() == shim.TypeList || sch.Type() == shim.TypeSet {
+			return sch.MaxItems() == 1
+		}
+		// If it's a map of resources then it's maxItems is 1
+		elem := sch.Elem()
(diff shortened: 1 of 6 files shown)


Zaid-Ajaj (approved):

Frassle:
Test failures are just the existing example failures, so merging this forward.

pkg/convert/tf_scopes.go
```diff
@@ -375,10 +375,18 @@ func (s *scopes) maxItemsOne(fullyQualifiedPath string) bool {
...
-	// If we have a shim schema use it's MaxItems
+	// If we have a shim schema use it's MaxItems and Type
 	sch := info.Schema
 	if sch != nil {
-		return sch.MaxItems() == 1
+		// If this is a list or set, check if it has a maxItems of 1.
+		if sch.Type() == shim.TypeList || sch.Type() == shim.TypeSet {
+			return sch.MaxItems() == 1
+		}
+		// If it's a map of resources then it's maxItems is 1
```
t0yv0, line 385: https://github.com/pulumi/pulumi-terraform-bridge/blob/f6e5260a88eab21043912e374bbafd5f42fd7f44/pkg/tfshim/shim.go#L146

This could be more helpful to refer to as a single-nested Terraform block.

MaxItems=1 introduces a displacement between Pulumi and Terraform representations. That is, under MaxItems=1 there is a collection in Terraform that gets mapped to a non-collection under Pulumi.

In this case, however, there is no collection, both Pulumi and Terraform representations have an object type, but due to historical limitations of the shim encoding it's saying TypeMap instead of something like TypeObject.

PR #231, opened 2024-11-26
Revert "Add package blocks to pcl conversion. (#221)"

Frassle:
This reverts commit ccb6e15d26933d24e89075b1b6cb85749f522f2a.

Fixes #224

PR #232, opened 2024-12-02
Reapply PCL package blocks

brandonpollack23:
Reverts pulumi/pulumi-converter-terraform#231 which was a revert of https://github.com/pulumi/pulumi-converter-terraform/commit/ccb6e15d26933d24e89075b1b6cb85749f522f2a

The goal is to try to integrate this without regressing #224


brandonpollack23:
change the provider parameterized name to just be the same as provider "name"
    Do a better check for if it is pulumiverse, use https://github.com/pulumi/pulumi/blob/master/pkg/util/plugin.go, and https://github.com/pulumi/ci-mgmt/blob/master/provider-ci/providers.json merged
    see if it works and split the initial block from the param block additions

brandonpollack23:
Moved to #239 and #240

PR #235, opened 2024-12-02
Test list and map objects as attributes

Frassle:
List and map objects can be assigned as attributes or blocks. We had only really been testing them as blocks, both in the complex_X tests and also the list_as_blocks and maps_as_blocks tests.

This changes the complex_X tests to instead use attribute syntax, to test that works.

I've updated the complex schema to split the one "inner_object" field into "inner_list_object" and "inner_map_object". This required updating a couple of other tests that were using the complex resource to refer to "inner_list_object" instead of "inner_object".

### DIFF
--- a/pkg/convert/tf_scopes.go
+++ b/pkg/convert/tf_scopes.go
@@ -348,7 +348,9 @@ func (s *scopes) isMap(fullyQualifiedPath string) *bool {
 	// If we have a shim schema use the type from that
 	sch := info.Schema
 	if sch != nil {
-		isMap := sch.Type() == shim.TypeMap
+		// This is only an actual _map_ if the elem isn't a resource
+		_, resource := sch.Elem().(shim.Resource)
+		isMap := sch.Type() == shim.TypeMap && !resource
 		return &isMap
 	}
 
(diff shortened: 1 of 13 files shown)


pkg/convert/tf_scopes.go
```diff
@@ -361,6 +363,26 @@ func (s *scopes) isMap(fullyQualifiedPath string) *bool {
 	return nil
 }
 
+// Given a fully typed path (e.g. data.simple_data_source.a_field) returns whether a_field is a resource object.
+func (s *scopes) isResource(fullyQualifiedPath string) bool {
+	info := s.getInfo(fullyQualifiedPath)
+
+	// This should only be called for attribute paths, so panic if this returned a resource
+	contract.Assertf(info.ResourceInfo == nil, "maxItemsOne called on a resource or data source")
+	contract.Assertf(info.DataSourceInfo == nil, "maxItemsOne called on a resource or data source")
```
lunaris, line 372: Should these be "must be called" or "not called"?

pkg/convert/tf_scopes.go
```diff
@@ -361,6 +363,26 @@ func (s *scopes) isMap(fullyQualifiedPath string) *bool {
...
+
+	// If we have a shim schema use it's MaxItems and Type
```
lunaris, line 374:
```suggestion
	// If we have a shim schema use its MaxItems and Type
```
:santa:

lunaris (approved):

pkg/convert/tf_scopes.go
```diff
@@ -348,7 +348,9 @@ func (s *scopes) isMap(fullyQualifiedPath string) *bool {
 	// If we have a shim schema use the type from that
 	sch := info.Schema
 	if sch != nil {
-		isMap := sch.Type() == shim.TypeMap
+		// This is only an actual _map_ if the elem isn't a resource
+		_, resource := sch.Elem().(shim.Resource)
+		isMap := sch.Type() == shim.TypeMap && !resource
```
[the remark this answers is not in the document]
Frassle: It's a double search for getInfo but that's probably neater

pkg/convert/tf_scopes.go
Frassle: must be

pkg/convert/tf_scopes.go
```diff
@@ -382,11 +404,6 @@ func (s *scopes) maxItemsOne(fullyQualifiedPath string) bool {
 		if sch.Type() == shim.TypeList || sch.Type() == shim.TypeSet {
 			return sch.MaxItems() == 1
 		}
-		// If it's a map of resources then it's maxItems is 1
-		elem := sch.Elem()
-		if _, isResource := elem.(shim.Resource); sch.Type() == shim.TypeMap && isResource {
-			return true
-		}
 	}
```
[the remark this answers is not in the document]
Frassle: It was wrong. maxItemsOne only applies to lists, it should never have applied to maps.

Issue #234, opened 2024-12-02
Apply package blocks to PCL


justinvp:
Now that PCL supports https://github.com/pulumi/pulumi/pull/17589, we should leverage this when converting.

PR #239, opened 2024-12-03
Add package block generation to the terraform converter.

brandonpollack23:
This is the first step in supporting package blocks.  The next step is
supporting parameterization and detecting if the package should be
bridge with terraform-bridge or not.

part of #234


pkg/convert/testdata/programs/required_providers/main.tf
```diff
@@ -0,0 +1,16 @@
+terraform {
+  required_providers {
+    random = {
+      version = "~> 3.5.1"
```
Frassle, line 4: Can we translate TF versions to Pulumi versions? I don't think so 1-2-1, but worth asking providers team about it.

pkg/convert/testdata/programs/required_providers/main.tf
```diff
@@ -0,0 +1,16 @@
...
+      source  = "hashicorp/random"
+    }
+    null = {
```
Frassle, line 7: Add a case here where the name doesn't match the source, e.g. like in https://developer.hashicorp.com/terraform/language/providers/requirements#handling-local-name-conflicts

pkg/convert/tf.go
```diff
@@ -3236,6 +3250,22 @@ func translateModuleSourceCode(
 
 		body := file.Body()
 
+		// Add package metadata to only the top level module.
+		// This should be true for only the entry point into the conversion, all
+		// recursive module translations should set this false.
+		// This guarantees that the package blocks are only written in one place (eg main.pp)
+		if topLevelModule {
+			for name, provider := range requiredProviders {
+				block, d := getPackageBlock(name, provider)
```
Frassle, line 3259: I think this needs to parse provider.Source() for the actual name, not just use the local name.

pkg/internal/shim/go.mod
Frassle, line 1: We're not using these shim files anymore for this PR? Can they just be removed.

pkg/convert/testdata/programs/required_providers/main.tf
brandonpollack23: No we can't really without a bunch of work.  I think we can leave it for now (it is unused and the version string from within the value payload is what is used to parameterize)

pkg/convert/testdata/programs/required_providers/main.tf
brandonpollack23: done.

pkg/internal/shim/go.mod
brandonpollack23: they are needed on the next change, moved them up

pkg/convert/tf.go
brandonpollack23: well we need the "name" to know what to name the package block, but I think you mean parse it to do the check if it is in pulumiverse or not in the next change?

Otherwise we should keep the name of the block the same as in tf, yes?

pkg/convert/testdata/programs/required_providers/pcl/main.pp
```diff
@@ -0,0 +1,12 @@
+package "different-name-null" {
+  baseProviderName = "different-name-null"
```
Frassle, line 2: This is why I said you need to look at .Source() at https://github.com/pulumi/pulumi-converter-terraform/pull/239/files#r1867179326. The base provider and package label here should be "null".

I'd also use a different provider than null, like http so that you don't have a package name conflict in this test.

Frassle (changes requested):

pkg/convert/testdata/programs/required_providers/pcl/main.pp
```diff
@@ -0,0 +1,12 @@
+package "different-name-null" {
```
Frassle, line 1: I think this need to be "null" for PCL as well. When the binder looks up package descriptors it does it based on this label, and the first part of the resource/invoke token. So if this is "different-name-null" nothing will ever use it.

Frassle (approved):

PR #240, opened 2024-12-03
Add paramaterization to package block generation

brandonpollack23:
Add parameterization to pulumi package blocks.

This adds parameterization for terraform providers which require dynamic
bridging (and do not have pulumiverse or official pulumi supported packages already).

Fixes: #234


brandonpollack23:
The failing test is a lie, that file isn't even changed in this PR...also the file itself matches what it says it doesn't.  My guess is this is another github cache problem.

pkg/convert/tf.go
```diff
@@ -3416,11 +3419,37 @@ func getPackageBlock(name string, prov *configs.RequiredProvider) (*hclwrite.Blo
...
+	packageNameParts := strings.Split(prov.Source, "/")
+	packageName := packageNameParts[len(packageNameParts)-1]
+	if isTerraformProvider(packageName) {
+		body.SetAttributeValue("baseProviderName", cty.StringVal("terraform-provider"))
+		body.SetAttributeValue("baseProviderVersion", cty.StringVal("0.3.0"))
+
+		var desiredVersion versions.Version
+		desiredVersion, diags = shim.FindTfPackageVersion(prov)
+
+		innerValue := fmt.Sprintf(`{"remote":{"url":"%s","version":"%s"}}`,
```
Frassle, line 3432: I wonder if we should depend on tf bridge for building this parameter value. It's simple now, but I know Ian has commented about potentially doing more pre-processing at parameterisation time which would change this value.
Maybe at least leave a comment here about that?

brandonpollack23: Comment added

pkg/convert/tf.go
```diff
@@ -3419,11 +3422,37 @@ func getPackageBlock(name string, prov *configs.RequiredProvider) (*hclwrite.Blo
...
+
+		var desiredVersion versions.Version
+		desiredVersion, diags = shim.FindTfPackageVersion(prov)
+
+		innerValue := fmt.Sprintf(`{"remote":{"url":"%s","version":"%s"}}`, 
+															prov.Source,
+															desiredVersion.String())
+		encoded := base64.StdEncoding.EncodeToString([]byte(innerValue))
+
+		paramBlock := hclwrite.NewBlock("parameterization", []string{"name", "version", "value"})
```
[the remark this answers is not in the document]
brandonpollack23: Yeah I thought they were like a schema thing but I'm realizing it only worked coincidentally now after this comment and talking w/ Fraser about it.

pkg/convert/tf.go
Frassle: I don't see the comment for this?

Frassle (approved):

PR #252, opened 2024-12-12
Add kubernetes to pulumiverse

brandonpollack23:
We have a kubernetes provider so we need to add it to our list in the converter of things _not_ to bridge to terraform.  As of now we'll end up with a terraform provider for k8s instead of using our native one.

This will fix that.


tgummerer:
This PR could benefit from a description describing why this is being done.

tgummerer (approved):
Thanks for adding the description!

PR #255, opened 2024-12-19
Release 1.0.21

brandonpollack23:
Push a release so that pulumi/pulumi can move forward

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 CHANGELOG
 =========
 
+## 1.0.21
+
+### Improvements
+
+- Change inferred resource names to pascal case
+- Add parameterization block to "package" blocks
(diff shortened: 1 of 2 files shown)


pkg/convert/pulumiverse.go
```diff
@@ -56,6 +56,7 @@ var pulumiSupportedProviders = []string{
 	"kafka",
 	"keycloak",
 	"kong",
+	"kubernetes",
```
tgummerer, line 59: This change seems unintentional here?  It should probably be in a different PR than the release PR.

brandonpollack23: This PR is just on top of that one so I don't submit them in the wrong order, just filter by latest commit :)

tgummerer (approved):

Issue #65, opened 2018-10-02
Support all TF built-in functions


pgavlin:
The complete list is [here](https://www.terraform.io/docs/configuration/interpolation.html#supported-built-in-functions); function-by-function status is below.

- [x] [abs(float)](https://www.terraform.io/docs/configuration/interpolation.html#abs-float-)
- [x] [basename(path)](https://www.terraform.io/docs/configuration/interpolation.html#basename-path-)
- [x] [base64decode(string)](https://www.terraform.io/docs/configuration/interpolation.html#base64decode-string-)
- [x] [base64encode(string)](https://www.terraform.io/docs/configuration/interpolation.html#base64encode-string-)
- [x] [base64gzip(string)](https://www.terraform.io/docs/configuration/interpolation.html#base64gzip-string-)
- [x] [base64sha256(string)](https://www.terraform.io/docs/configuration/interpolation.html#base64sha256-string-)
- [x] [base64sha512(string)](https://www.terraform.io/docs/configuration/interpolation.html#base64sha512-string-)
- [x] [bcrypt(password, cost)](https://www.terraform.io/docs/configuration/interpolation.html#bcrypt-password-cost-)
- [x] pulumi/tf2pulumi#263
- [x] [chomp(string)](https://www.terraform.io/docs/configuration/interpolation.html#chomp-string-)
- [x] [chunklist(list, size)](https://www.terraform.io/docs/configuration/interpolation.html#chunklist-list-size-)
- [x] [cidrhost(iprange, hostnum)](https://www.terraform.io/docs/configuration/interpolation.html#cidrhost-iprange-hostnum-)
- [x] [cidrnetmask(iprange)](https://www.terraform.io/docs/configuration/interpolation.html#cidrnetmask-iprange-)
- [x] [cidrsubnet(iprange, newbits, netnum)](https://www.terraform.io/docs/configuration/interpolation.html#cidrsubnet-iprange-newbits-netnum-)
- [x] [coalesce(string1, string2, ...)](https://www.terraform.io/docs/configuration/interpolation.html#coalesce-string1-string2-)
- [x] [coalescelist(list1, list2, ...)](https://www.terraform.io/docs/configuration/interpolation.html#coalescelist-list1-list2-)
- [x] [compact(list)](https://www.terraform.io/docs/configuration/interpolation.html#compact-list-)
[...]

PR #266, opened 2025-01-31
Re-enabling compact for converter

brandonpollack23:
Compact is disabled but in local testing trying to make convert more powerful this seems to work.

Reenabling for now because it is not working with or without it anyway

Blocked by pulumi/pulumi-std#76
Blocked by pulumi/pulumi-std#77

Related #65
Fixes pulumi/pulumi#18406


brandonpollack23:
Test is failing not ready for reveiw

PR #269, opened 2025-01-31
enable coalesce now that it is in pulumi-std

brandonpollack23:
blocked by pulumi/pulumi-std#78


brandonpollack23:
After this is merged we need to push a release

PR #270, opened 2025-01-31
Enable merge invoke.

brandonpollack23:
Fixes https://github.com/pulumi/pulumi/issues/18356

Related https://github.com/pulumi/pulumi-converter-terraform/issues/65


Frassle (changes requested):
Again, merge is generic it should not even be in pulumi-std it throws away all typing information. This should be a pcl intrinsic.

brandonpollack23:
After looking closer at this I think it isn't generic it's dynamic.

https://developer.hashicorp.com/terraform/language/functions/merge

These objects in the example are dynamic, they can be any kind of map or object, so in that case there is no generic that will necessarily satisfy any requirment unless you manually define a type that happens to be the appropriate merge of two other types

Frassle (approved):
Talking to Will we'll try going with these std "any" variants for now, I suspect we're going to have to be smarter about types to get good conversions though, but we can test that with seeing how some more top modules convert

PR #271, opened 2025-01-31
Enable flatten

### LINKED ISSUE #138: Support flatten
## Hello!

## Issue details

Rare but for completeness sake can be good to support the `flatten` function. Only 1 AWS example is blocked by this:

```terraform
    resource "aws_eks_node_group" "example" {
      cluster_name    = "example"
      node_group_name = "example"

      # ... other configuration ...
    }

    resource "aws_autoscaling_group_tag" "example" {
      for_each = toset(
        [for asg in flatten(
          [for resources in aws_eks_node_group.example.resources : resources.autoscaling_groups]
        ) : asg.name]
      )

      autoscaling_group_name = each.value

      tag {
        key   = "k8s.io/cluster-autoscaler/node-template/label/eks.amazonaws.com/capacityType"
        value = "SPOT"

        propagate_at_launch = false
      }
    }

```

brandonpollack23:
Fixes #138 

Fixes pulumi/pulumi#18355

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -4,6 +4,7 @@
 - Enable jsondecode which is already in pulumi-std
 - Enable lookup which is already in pulumi-std
 - Enable merge which is already in pulumi-std
+- Enable flatten which is already in pulumi-std
 
 ### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


Frassle (changes requested):
I don't think flatten should be in pulumi-std, you lose all the type information if you go through that invoke (it's just `list[any] -> list[any]`).

We should probably remove it from std, and then add the pcl intrinsic to go flatten generically.

brandonpollack23:
same thing here as #270 , lists already are list[any], they are not generically typed

Frassle (approved):

PR #273, opened 2025-02-03
Enable lookup, it is part of pulumi-std.

brandonpollack23:
Related #65
Fixes #pulumi/pulumi#18416


Frassle:
Again, I don't think this should be in std. We're going to re-type every `lookup` as any by doing this and it's a trivial generic function.

PR #277, opened 2025-02-06
Remove providers that are now dynamically bridged from pulumiverse.go

guineveresaenger:
Related to https://github.com/pulumi/home/issues/3876.

Removes the following providers as no longer statically bridged and maintained by Pulumi:

- Civo
- RKE
- Libvirt
- Local


guineveresaenger:
Test failures seem due to a failed publish run of pulumi-std.

Frassle (approved):

PR #279, opened 2025-02-10
Bump the test schema for `pulumi-std`

lunaris:
This repository uses a vendored copy of the schema for the `pulumi-std` provider in its tests. This change bumps this schema to the latest released copy (2.1.0), ahead of changing the converter to use some new functionality in this version. It also tweaks the linked version of `pulumi/terraform`, since it appears that the referenced v1.4.0 doesn't exist as a tag in that repository.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/pulumi/pulumi-terraform-bridge/v3 v3.96.0
 	github.com/pulumi/pulumi/pkg/v3 v3.142.0
 	github.com/pulumi/pulumi/sdk/v3 v3.142.0
-	github.com/pulumi/terraform v1.4.0
+	github.com/pulumi/terraform v0.12.1-0.20230322133416-a268cd0892c9
 	github.com/spf13/afero v1.11.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.9.0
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/pulumi/pulumi-terraform-bridge/v3 v3.96.0
 	github.com/pulumi/pulumi/pkg/v3 v3.142.0
 	github.com/pulumi/pulumi/sdk/v3 v3.142.0
-	github.com/pulumi/terraform v1.4.0
+	github.com/pulumi/terraform v0.12.1-0.20230322133416-a268cd0892c9
```
tgummerer, line 19: This is confusing, 1.4.0 indeed doesn't exist, so I'm confused how this ever worked.  But also where does `a268cd...` come from?  I don't see that as the latest commit in any of the branches in that repo :thinking:

lunaris: https://github.com/pulumi/terraform/commit/a268cd0892c930d7881d7e8158a14ec146dfcaee -- confusingly, the default branch in the GH UI is `pulumi-master`, but the most up-to-date (and used) is `main`. Even so, this commit is now 3 behind the `HEAD` of that branch, so we should probably bump this separately. This version is a revert to what was there before 1.4.0.

tgummerer: Thanks for the explanation. It would be good to know how the v1.4.0 tag disappeared in the first place, since clearly things seemed to be working :thinking: But this looks reasonable to me then.

PR #281, opened 2025-02-10
Implement the `compact` function

lunaris:
This should be implemented in the newly released `pulumi-std` 2.1.0, so we can start generating invoke calls for it.

Fixes pulumi/pulumi#18406

### DIFF
--- a/pkg/convert/tf.go
+++ b/pkg/convert/tf.go
@@ -493,21 +493,11 @@ var tfFunctionStd = map[string]struct {
 		output:    "result",
 		paramArgs: true,
 	},
-	/* Currently failing due to: cannot assign expression of type { input: (string, string, string, null,
-	string) } to location of type { input: list(output(string) | string) | output(list(string)) } | output({
-	input: list(string) }):
-	*/
-	//
-	//	on main.pp line 263:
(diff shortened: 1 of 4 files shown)


brandonpollack23 (approved):

pkg/convert/tf.go
```diff
@@ -493,21 +493,11 @@ var tfFunctionStd = map[string]struct {
...
-	//"compact": {
-	//	token:  "std:index:compact",
-	//	inputs: []string{"input"},
-	//	output: "result",
-	//},
+	"compact": {
+		token:  "std:index:compact",
+		inputs: []string{"input"},
+		output: "result",
+	},
```
brandonpollack23, line 500: doesnt this need paramArgs?

brandonpollack23 (changes requested):

pkg/convert/tf.go
lunaris: I'm not sure it does -- https://developer.hashicorp.com/terraform/language/functions/compact apparently it just takes "a list of strings", so in that sense it's "already `paramArgs`-d". See also the existing test case, that passes a single list of strings explicitly.

brandonpollack23 (approved):

PR #286, opened 2025-02-10
Use `format` from `pulumi-std`

lunaris:
Fixes https://github.com/pulumi/pulumi/issues/18521


pkg/convert/testdata/programs/builtin_functions/main.tf
```diff
@@ -406,26 +406,20 @@ output "funcFormat2" {
...
-output "funcFormat6" {
+output "funcFormat5" {
   value = format("%#v", 1)
 }
-output "funcFormat7" {
+output "funcFormat6" {
   value = format("%#v", {a = 1})
 }
-output "funcFormat8" {
+output "funcFormat7" {
```
brandonpollack23, line 420: I'd feel a little better if we had one example with more than one format argument as well to be sure the packing you added works for that

brandonpollack23 (changes requested):

pkg/convert/testdata/programs/builtin_functions/main.tf
lunaris: Good call -- done :+1:

brandonpollack23 (approved):

PR #296, opened 2025-02-12
Use toset from pulumi-std

### LINKED ISSUE #137: Support toset
## Hello!

## Issue details

Lack of `toset` function support blocks a handful (4) AWS examples. The most prominent one is:

```hcl

    resource "aws_eks_node_group" "example" {
      cluster_name    = "example"
      node_group_name = "example"

      # ... other configuration ...
    }

    resource "aws_autoscaling_group_tag" "example" {
      for_each = toset(
        [for asg in flatten(
          [for resources in aws_eks_node_group.example.resources : resources.autoscaling_groups]
        ) : asg.name]
      )

      autoscaling_group_name = each.value

      tag {
        key   = "k8s.io/cluster-autoscaler/node-template/label/eks.amazonaws.com/capacityType"
        value = "SPOT"

        propagate_at_launch = false
      }
    }

```

Sets are represented as lists in Pulumi but this function deduplicates at runtime. Perhaps we can get by with a non-deduplicating implementation for starters, since the provider will deduplicate anyway.

daveset:
This enables the `toset` function from pulumi-std

Ran `gofumpt -w pkg` which accounts for the whitespace changes.

Fixes #137 

Note: when I run `make lint` locally I get a bunch of failures unrelated to my changes. Note that the Github action for lint is using an invalid make target and does not fail the build on lint errors - see https://github.com/pulumi/pulumi-converter-terraform/blob/7741cabca70d832cf2a8d15a66c5538f44fa2c98/.github/workflows/stage-lint.yml#L38: (`make lint-golang || true` should be `make lint`).


brandonpollack23:
Yay welcome Dave!

Go ahead and file an issue to fix that presubmit action and fix it if you have a second (I think the issue can go to pulumi/pulumi, it seems we're aggregating things there now)

pkg/internal/shim/go.sum
```diff
@@ -425,6 +425,7 @@ github.com/pkg/errors v0.9.1/go.mod h1:bwawxfHBFNV+L2hUp1rHADufV3IMtnDRdf1r5NINE
 github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
 github.com/prometheus/client_model v0.0.0-20190812154241-14fe0d1b01d4/go.mod h1:xMI15A0UPsDsEKsMN9yxemIoYk6Tm2C1GtYGdfGttqA=
+github.com/pulumi/terraform v0.12.1-0.20230322133416-a268cd0892c9/go.mod h1:w029Faz4rGcWSr4gPHWjlzU7CdvbxwQNjgiYWBC0FzU=
```
brandonpollack23, line 428: did you also run go mod tidy?

brandonpollack23 (approved):
In the future let's keep whitespace changes to their own PR.  Generally folks aren't a fan of them b/c they muck up git blame (though I think tidying is good as well so YMMV).

Check comments before merging

pkg/internal/shim/go.sum
daveset: Hi Brandon. Thanks for the review!

I do think I ran `go mod tidy` either via `make ensure` or goland IDE. The go.sum entry reflects a change that was added to `pkg/internal/shim/go.mod` on 2/10/25 but was not updated along with that PR.

I am happy to back that out if you don't want to pull that change in. 

Same with the whitespace. TBH I'm not sure if that was from running gofumpt or from the IDE running its formatter.

lunaris:
Hi @daveset. Thanks again for raising this PR, and bringing to our attention the linting issues. I've since merged some changes that should mean linting is working again, and as a result, the repo is also `gofumpt`'d. If you're able to do (hopefully) one last rebase on `main` now, all this should get picked up and we can run the tests on your branch and get this merged :pray: Cheers!

daveset:
Thanks @lunaris! Happy to help - my conversion was failing on this `toset` and the unimplemented `try` that has an open draft PR.  It looks like you all have been busy with similar updates.  

Rebased and test/lint/gofumpt all works locally so it should be ready to go.

pkg/internal/shim/go.sum
brandonpollack23: I dont mind the formatting in this one tbh, but I've actually gotten the feedback enough (due to git blame) to turn off autoformatting for some file types.  For me it was that i had gofumpt_lines or some such set where it would try to split lines that are too long and that isnt the default.  YMMV due to git blame preservatoin requests. With go mod tidy, this one is fine but in the future I think it should be called out or its own PR

daveset: I have had the same experience with formatting, mostly when it comes to switching between different projects that each have their own unique requirements. I agree about the go mod tidy change - normally I wouldn't mix that in but I clearly missed it.

Thank you for getting this merged, along with all the other recent updates going in here. The last one blocking the conversion i'm trying to run is `try`, and I know you have that in progress in another PR.

brandonpollack23:
I hope to unblock you soon, we're hard at work on it.  We may need to
massage `try` some more but we hope to get you unblocked soon!

PR #304, opened 2025-02-18
Add extra warning to help users of advanced resource lifecycles be aware of differences

brandonpollack23:
Pulumi creates resources before destroy by default unlike TF.  There is
a flag to do that in TF, and when it is set, chances are the user may
not be aware pulumi does this, or worse they expect opposite behavior.
Just in case, we emit a warning since they're using lifecycle flags
anyway, and they may wish to know either:

* Why it didn't convert as written
* That their other resources they expect that behaviour need to be
  modified.


lunaris (approved):
:ok_hand:

brandonpollack23:
This fixed pulumi/pulumi#18564

PR #305, opened 2025-02-18
Fix failing tests from unaccepted changes

brandonpollack23:
Followon to PR #304 to fix tests

### DIFF
--- a/pkg/convert/testdata/states/range/import.json
+++ b/pkg/convert/testdata/states/range/import.json
@@ -1,8 +1,8 @@
 [
   {
     "Type": "simple:index:resource",
-    "Name": "a_resource-hello",
-    "ID": "abc123",
+    "Name": "a_resource-goodbye",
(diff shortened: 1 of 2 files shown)


pkg/convert/testdata/states/range/import.json
```diff
@@ -1,8 +1,8 @@
 [
   {
     "Type": "simple:index:resource",
-    "Name": "a_resource-hello",
-    "ID": "abc123",
```
lunaris, line 5: I think you can remove this file -- this test is non-deterministic. I've been meaning to take a look at it (probably a map in Go or something).

lunaris (approved):

pkg/convert/testdata/states/range/import.json
brandonpollack23: I've run into a few of these as well and usually end up reading into a key sorted list to fix it

PR #306, opened 2025-02-21
Add ability to rename provider configurations for conversions.

brandonpollack23:
Previously, we would not rename certain provider configuration keys,
which results in errors after conversion.  This adds a map to add these
cases as we find them, starting with google to GCP.

Fixes pulumi/pulumi-converter-terraform#307


Frassle:
Yeh this definitely feels like it should be using the mapping information generally rather than having a hardcoded lookup to decided it.

Issue #228, opened 2024-11-26
Attempting conversion containing `for_each` with primitives fails


smithrobs:
### What happened?

The `pulumi convert --from terraform --language python` command does not work for terraform for-loops that have primitives. For example `for_each = var.enable ? [1]  : []` which we typically do for dynamic blocks such as S3 lifecycle rules which has to be defined in a single resource ([ref](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_lifecycle_configuration)). The error message is:
```
error: main.pp:7,40-75: cannot traverse value of type number;
error: main.pp:7,40-75: cannot traverse value of type union(cty.NumberIntVal(1), none);
error: could not generate output program
```

### Example

```hcl
resource "aws_s3_bucket_lifecycle_configuration" "this" {
  count  = length(var.expirations) > 0 || length(var.transitions) > 0 || length(var.noncurrent_version_transitions) > 0 ? 1 : 0
  bucket = aws_s3_bucket.this.id
  
  # There are some other dynamic rule blocks i have omitted

  dynamic "rule" {
    for_each = var.transitions
    content {
      status = try(rule.value.enabled, true) ? "Enabled" : "Disabled"
      id     = try(rule.value.id, md5("transitions${jsonencode(rule.value)}"))
      dynamic "abort_incomplete_multipart_upload" {
        # following line will fail to convert
        for_each = try(rule.value.abort_incomplete_multipart_upload_days, null) != null ? [1] : []
        content {
          days_after_initiation = rule.value.abort_incomplete_multipart_upload_days
        }
      }
      filter {
        prefix = try(rule.value.prefix, "")
        dynamic "tag" {
          for_each = try(jsondecode(rule.value.tags), [])
          content {
            key   = try(tag.key, null)
            value = try(tag.value, null)
          }
        }
      }
      transition {
        date          = try(rule.value.date, null)
        days          = try(rule.value.days, null)
        storage_class = rule.value.storage_class
      }
    }
  }
}
```

### Output of `pulumi about`

Not available; assume v3.141.0

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

joeduffy:
@brandonpollack23 Would love it if you could take a look at this. Sounds like it could be impactful for the customer @smithrobs is working with!

brandonpollack23:
Taking a look!

[...]

PR #310, opened 2025-02-26
Fix nested dynamic blocks shadowing names

brandonpollack23:
This was causing names that needed to be accessible to be shadowed,
resulting in failed PCL binding after conversion see https://github.com/pulumi/pulumi-converter-terraform/issues/228

Fixes https://github.com/pulumi/pulumi/issues/18718


pkg/convert/tf_scopes.go
```diff
@@ -147,6 +147,60 @@ func (s *scopes) generateUniqueName(name, prefix, suffix string) string {
 	}
 }
 
+// addNestedScopeUniqueName adds a name to the current scope, making it unqiue
+// if needed.  Returns a function to cleanup any root modificaitons.
+func (s *scopes) addNestedScopeUniqueName(name, prefix, suffix string) (string, func()) {
```
brandonpollack23, line 152: let the record show I'm not in _love_ with this solution, but I couldnt see a more obvious way to do this kind of shadowing avoidance and still maintain names as being used etc.  The deferred cleanup return is needed so we don't fill the root with things permanently and always increment "entry", but we don't always have a locals scope available

brandonpollack23:
yay passing

pkg/convert/testdata/programs/dynamic_nested_for_each/main.tf
```diff
@@ -0,0 +1,52 @@
...
+  default = []
+}
+
+resource "blocks_resource" "a_resource" {
+    count = length(listvar) > 0 ? 1 : 0
+    dynamic "a_list_of_resources" {
+        for_each = var.dynvar
+
+        content {
+            inner_string = try(a_list_of_resources.value.inner_value, null) != null ? "TrySuccess" : "TryFail"
```
Frassle, line 21: Do we need to conflate "try" into this test? This is just supposed to be around nested dynamics, lets not make it more complex with try expressions as well.

brandonpollack23: you're right, this isn't needed

pkg/convert/testdata/mappings/blocks.json
```diff
@@ -43,6 +43,18 @@
                             "inner_string": {
                                 "type": 4,
                                 "optional": true
+                            },
+                            "inner_dynamic_resources": {
```
Frassle, line 47: Nothing about this is strictly dynamic? Like we're happening to use it in a dynamic block but we could use this for any other nested resource tests as well

pkg/convert/tf_scopes.go
```diff
@@ -147,6 +147,60 @@ func (s *scopes) generateUniqueName(name, prefix, suffix string) string {
 	}
 }
 
+// addNestedScopeUniqueName adds a name to the current scope, making it unqiue
+// if needed.  Returns a function to cleanup any root modifications.
+func (s *scopes) addNestedScopeUniqueName(name, prefix, suffix string) (string, func()) {
+	addAndReturn := func(name string) (string, func()) {
+		cleanup := func() {}
+		if len(s.locals) == 0 {
```
Frassle, line 155: Why are we only adding to locals if it's non-empty? for loop expression always add to locals, shouldn't dynamic blocks always add to locals as well?

pkg/convert/tf_scopes.go
```diff
@@ -147,6 +147,60 @@ func (s *scopes) generateUniqueName(name, prefix, suffix string) string {
...
+			s.roots[name] = PathInfo{Name: name}
+			cleanup = func() { delete(s.roots, name) }
+		} else {
+			s.locals[len(s.locals)-1][name] = name
```
Frassle, line 159: Don't we still need to cleanup the local? If we add a new local it still needs to fall out of scope after the dynamic expression

pkg/convert/testdata/mappings/blocks.json
brandonpollack23: no i just wanted to signal what its for but ill rename it

pkg/convert/tf_scopes.go
brandonpollack23: I'm not sure why I didn't just use append here, this is odd.

let me change it and I think you're right it should work fine and be way simpler

pkg/convert/tf_scopes.go
brandonpollack23: This is why I did the gross root thing, and why jsut pushing a scope is better

pkg/convert/tf_scopes.go
brandonpollack23: yeah this is so much better, what was I thinking?

brandonpollack23:
ping

Frassle (approved):
Looks ok, I think we could simplify some bits around scopes now but this can go in as is while we think about that.

PR #311, opened 2025-02-26
upgrade pulumi/pulumi to latest

tgummerer:
So we can generate dotnet 8.0 projects correctly.


Frassle:
This shouldn't need an update... the codegen is invoked by the engine not by the converter :|

tgummerer:
Hrm I guess we need to figure out why it seems to emit an old .net version then :/  Gonna close this PR out in the meantime.

PR #319, opened 2025-03-31
Allow mapper package hint when a provider was not found in the registry

guineveresaenger:
This pull request allows for the converter to convert irrespective of whether a real provider was found on the Terraform registry.

This was discovered when running tests in the terraform bridge using a partially mocked provider.

### DIFF
--- a/pkg/convert/info.go
+++ b/pkg/convert/info.go
@@ -68,7 +68,7 @@ func (s *mapperProviderInfoSource) GetProviderInfo(
 	}
 
 	var hint *convert.MapperPackageHint
-	mappingMessage := fmt.Sprintf("could not find mapping information for provider %s;"+
+	mappingMessage := fmt.Sprintf("could not find mapping information for provider %s; "+
 		"try installing a pulumi plugin that supports this terraform provider",
 		tfProvider)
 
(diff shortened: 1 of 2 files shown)


brandonpollack23:
I'm not sure this is such a good idea, what does it mean to convert something with a nonexistent provider? Can we modify the test to use a local provider instead or some such?

pkg/convert/info.go
```diff
@@ -75,22 +75,24 @@ func (s *mapperProviderInfoSource) GetProviderInfo(
 	if isTerraformProvider(pulumiProvider) && requiredProvider != nil {
 		tfVersion, diags := shim.FindTfPackageVersion(requiredProvider)
 		if diags.HasErrors() {
-			return nil, fmt.Errorf("could not find version for terraform provider %s: %v", tfProvider, diags)
```
Frassle, line 78: We should probably add a warning diagnostic here, but I think it's fine to still try and do the conversion without a hint.

guineveresaenger: I'm happy to pass a warning, but what do you suggest to use for "without a hint"? Perhaps I'm missing something about how hints are used but I was going for "most obvious default".

Frassle: Something like:
"Could not find TF version for '<requiredProvider>' assuming a Pulumi provider '<pulumiProvider>' exists."

guineveresaenger: ohhhh you just want to add a log line explaining we're going with the assumption proposed. Gotcha, thank you!

Frassle (approved):

PR #322, opened 2025-04-22
Add newline after heredoc close delimiter so that tuplecons expressions etc parse correctly

brandonpollack23:
Fixes pulumi/pulumi#19217

### DIFF
--- a/pkg/convert/testdata/programs/expressions/pcl/main.pp
+++ b/pkg/convert/testdata/programs/expressions/pcl/main.pp
@@ -92,6 +92,7 @@ output "heredoc" {
 This is also a template.
 So we can output the key again ${aKey}
 END
+
 }
 
 output "heredocWithHyphen" {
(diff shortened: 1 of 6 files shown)


pkg/convert/testdata/programs/expressions/pcl/main.pp
```diff
@@ -91,14 +91,18 @@
   value = <<END
 This is also a template.
 So we can output the key again ${aKey}
+
```
Frassle, line 94: This looks wrong, you've added an extra newline to each string

pkg/convert/tf.go
```diff
@@ -1352,8 +1352,10 @@ func convertTemplateExpr(state *convertState,
 					if tok.Type == hclsyntax.TokenQuotedLit && isHereDoc {
 						splitStrings := strings.Split(string(tok.Bytes), "\\n")
 						for i, splitString := range splitStrings {
-							strtoksWithNewlines = append(strtoksWithNewlines, makeToken(hclsyntax.TokenStringLit, splitString))
-							if i < len(splitStrings)-1 && partIndex != len(expr.Parts)-1 {
+							if splitString != "" {
```
Frassle, line 1355: How does this handle strings like:
```
<<END

This is also a template.

So we can output the key again ${local.a_key}

END
```

Feels like we should add that to the expressions test file and check it doesn't _lose_ new lines

brandonpollack23: good idea, in fact I think there is a bug here rn

Frassle (approved):

PR #324, opened 2025-05-06
Support converting remote modules to resources parameterized by terraform-module

### LINKED ISSUE #323: Permit targeting modules for direct execution
## Hello!

## Issue details

Permit targeting modules for direct execution.

@Zaid-Ajaj let me know if this is a duplicate, per our conversation you had some prototyping work to make this work as well as let users select which modules should be mapped to direct execution and which ones can continue being converted recursively.

Zaid-Ajaj:
### Descriptions

This PR implements support for converting module calls into (parameterized) resources backed by the `pulumi-terraform-module` provider when annotated with `// @sandbox <packageName>` this way, we no longer to download and convert the module sources, instead we convert the module call into a resource block in PCL and add a package block with `package <packageName> { ... }` that includes the parameterization details of the sandboxed module. 

When a module call is annotated with `// @sandbox <packageName>`, other module calls with the same source will be considered annotated/sandboxed with the same `<packageName>` such that the users do need to repeat the annotation.

See `sandbox_multiple_registry_modules` test for examples.

~~The test is marked `partial` because the test package loader doesn't handle loading parameterized packages~~

EDIT: the test is no longer partial and it type-checks

### Implementation details

We keep track of the module calls that are annotated with `// @sandbox <packageName>` in the conversion state and emit the associated package blocks at the top level packages PCL file. Before we translate module calls, we check if they are sandboxed, in which case we emit a resource block instead of downloading the module sources. 

When converting inputs of sandboxed modules, we maintain their casing, as per the convention of `pulumi-terraform-module` so we don't camelCase the input fields of these modules. Similarly, when referencing outputs of sandboxed modules, we keep the casing the output fields. 

Fixes #323


t0yv0:
I'd like to bikeshed the name a bit.

> In a nutshell, "sandboxing" in its general sense refers to a controlled environment for experimentation or testing, often used in computer security and software development.

The connotation here for experimentation/testing or perhaps security guarantees is a bit unhelpful.


AI:

```


Here are some alternative names for the feature and markers you can consider:

1. *@directExecute*: Suggests that the module will be executed directly without translation.

2. *@bypassTranslate*: Highlights that the translation process is bypassed in favor of direct execution.

3. *@inlineRun*: Implies running the module "inline" or directly without translation.

4. *@moduleProxy*: Communicates that the module call acts as a proxy, running as is.

5. *@deferConversion*: Indicates delaying the conversion process in favor of direct execution.

6. *@nativeCall*: Suggests that calls are made natively without modification.

7. *@resourceWrap*: Highlights that the module call is wrapped as a resource.

8. *@autoPackage*: Emphasizes automatic packaging or parameterization.

9. *@execBlock*: Implies executing as a block without conversion.

10. *@packageLink*: Indicates that the module is linked to a package for direct execution. 

These options aim to convey the idea of executing or utilizing modules directly while minimizing translation.


```

t0yv0:
Perhaps:

```
// @pulumi-terraform-module <packageName>
```

A bit more verbose but what I like about this is it's exclusively clear this is a pulumi annotation, and it has something that googles easily to finding more about this.

pkg/convert/testdata/programs/sandbox_multiple_registry_modules/pcl/main.pp
```diff
@@ -0,0 +1,30 @@
+package "subnets" {
+  baseProviderName    = "terraform-module"
+  baseProviderVersion = "0.1.3"
+  parameterization {
+    name    = "subnets"
+    version = "1.0.0"
+    value   = "eyJtb2R1bGUiOiJyZWdpc3RyeS50ZXJyYWZvcm0uaW8vaGFzaGljb3JwL3N1Ym5ldHMvY2lkciIsInBhY2thZ2VOYW1lIjoic3VibmV0cyIsInZlcnNpb24iOiIxLjAuMCJ9"
```
t0yv0, line 7: I'm still miffed this is unreadable. Does PCL have functions for base64-encoding? So we could splice a call to base64-encode a readable structure?

pkg/convert/tf.go
```diff
@@ -1703,10 +1715,34 @@ func rewriteTraversal(
 	return hclwrite.TokensForTraversal(newTraversal)
 }
 
+func isModuleRef(expr *hclsyntax.ScopeTraversalExpr) (string, bool) {
+	root := expr.Traversal.RootName()
+	if root != "module" {
+		return "", false
+	}
+	if attr, ok := expr.Traversal[1].(hcl.TraverseAttr); ok {
```
t0yv0, line 1723: Is [1] guaranteed to exist or needs a bounds check?

pkg/convert/tf.go
```diff
@@ -2657,6 +2699,54 @@ See https://www.pulumi.com/docs/iac/concepts/options/deletebeforereplace/ for de
...
+	moduleCall := sandboxedModule.moduleCall
+	path := "module." + moduleCall.Name
+	pulumiName := scopes.roots[path].Name
+
+	token := sandboxedModule.packageName + ":index:Module"
+	labels := []string{pulumiName, token}
+	block := hclwrite.NewBlock("resource", labels)
+	blockBody := block.Body()
+
+	// Does this resource have a count? If so set the "range" attribute
```
t0yv0, line 2717: Interesting is there a test that demonstrates this? You can call a module with a count?

pkg/convert/tf.go
```diff
@@ -2657,6 +2699,54 @@ See https://www.pulumi.com/docs/iac/concepts/options/deletebeforereplace/ for de
...
+	state.disableRewritingObjectKeys(func() {
+		// sandboxed modules keep their inputs and outputs without rewriting
+		moduleArgs := convertBody(state, scopes, path, moduleCall.Config)
+		for _, arg := range moduleArgs {
+			blockBody.SetAttributeRaw(arg.Name, arg.Value)
+		}
+	})
+
+	// Clear any index we set
+	scopes.countIndex = nil
```
t0yv0, line 2743: Not familiar with this code yet so noticing the mutability as unfortunate.

pkg/convert/testdata/programs/sandbox_registry_module/pcl/main.pp
```diff
@@ -0,0 +1,24 @@
+package "subnets" {
+  baseProviderName    = "terraform-module"
+  baseProviderVersion = "0.1.3"
+  parameterization {
+    name    = "subnets"
+    version = "1.0.0"
```
t0yv0, line 6: This is preserved from main.tf `version = "1.0.0"` right?

pkg/convert/tf.go
```diff
@@ -2869,10 +2960,74 @@ func translateRemoteModule(
...
+func moduleFromRemoteRegistry(moduleCall *configs.ModuleCall) (addrs.ModuleSourceRegistry, bool) {
+	switch source := moduleCall.SourceAddr.(type) {
+	case addrs.ModuleSourceRegistry:
+		return source, true
+	default:
+		return addrs.ModuleSourceRegistry{}, false
+	}
+}
+
+func resolveRemoteRegistryModule(moduleCall *configs.ModuleCall) (addrs.ModuleSourceRegistry, *version.Version, error) {
```
t0yv0, line 2993: This sounds a little fishy but I think I understand the need - it's trying to compute the concrete parameterization that requires a concrete version which we might not have if the module call did not specify any concrete version. So it's trying to find one. To some extent this is duplicating the functionality that terraform-module itself implements: that is when we say `pulumi package add terraform-module <x>` the "args" are resolved to parameterization including presumably this version selection pass. It would be ideal if this logic could live in one place. Perhaps take a TODO here? We could share code with the pulumi-terraform-module codebase or else we could find a way to dynamically talk to it, e.g. call `pulumi package add` and then figure out the resulting parameterization.

pkg/convert/tf.go
```diff
@@ -2869,10 +2960,74 @@ func translateRemoteModule(
...
+				packageName := parts[i+1]
+				return packageName, true
+			}
+		}
+	}
+
+	return "", false
+}
+
+func moduleFromRemoteRegistry(moduleCall *configs.ModuleCall) (addrs.ModuleSourceRegistry, bool) {
```
t0yv0, line 2984: Why does the code needs to branch on this bool returned here? What's different for non-remote modules?

pkg/convert/tf.go
```diff
@@ -3041,17 +3198,41 @@ func translateModuleSourceCode(
...
+					state.sandboxedModuleNames[moduleCall.Name] = moduleName
+					sandboxedModules[key] = &SandboxModule{
+						packageName: packageName,
+						moduleCall:  moduleCall,
+					}
+				}
+			}
+
+			// check if this module has been seen before. If it has, we don't need to
+			// translate it again.
```
t0yv0, line 3232: Can you explain a little what this short-circuits? I thought perhaps every module call needs to be translated fresh even if it maps to the similar version and such it's still a separate instance, or am I misunderstanding?

pkg/convert/tf.go
Zaid-Ajaj: Yes, traversal expressions in hcl always start with the type of the thing we are traversing e.g. `module`, `local`, `resource` etc then the name reference. For example `module.example_module` is a traversal that accesses the `example_module` so it does not require a bounds check

pkg/convert/tf.go
Zaid-Ajaj: >  You can call a module with a count?

I believe you can, both with `count` and `for_each` to instantiate multiple instances of the thing. I can add an example to the test

pkg/convert/testdata/programs/sandbox_registry_module/pcl/main.pp
Zaid-Ajaj: Yes

pkg/convert/tf.go
Zaid-Ajaj: I will add a TODO ✅

pkg/convert/tf.go
Zaid-Ajaj: I think this comment is for modules that aren't using `pulumi-terraform-module` so that are not downloaded and converted again. I will re-adjust it here

pkg/convert/tf.go
Zaid-Ajaj: We don't handle non-remote modules i.e. local with this implementation

pkg/convert/testdata/programs/sandbox_multiple_registry_modules/pcl/main.pp
t0yv0: >  so the closest way to represent that in text is base64-encoded string.

The extend to which we need to care I think is just the extent to which this is visible. Possibly PCL is not visible to the casual user converting to TypeScript so it is just an implementation detail then we do not care (just could be easier to debug if it was readable). I guess in the generated code the parameter value is not written anywhere yet right, at least at the moment? It is written/embedded into the generated SDK. There was some talk about writing it into Pulumi.yaml or similar, which would be another point observable by the user. It is those locations where users would "see" it that we could consider making more legible.

I guess we should punt on it for now unless it's really trivial here.

pkg/convert/tf.go
t0yv0: Nice ty !

pkg/convert/tf.go
t0yv0: Why not? That is kind of essential. That is our most important use case for enterprises trying to transition legacy modules maintained locally. What would it take?

pkg/convert/tf.go
t0yv0: Ah ok sorry I should have read closer. So some pre-existing behavior not quite affected.

pkg/convert/testdata/programs/sandbox_multiple_registry_modules/pcl/main.pp
Zaid-Ajaj: >I guess in the generated code the parameter value is not written anywhere yet right, at least at the moment? It is written/embedded into the generated SDK.

That's right, it is only embedded in the generated parameterized SDK

t0yv0:
We can probably ship this as is but I'd love to understand the local module bit. I think that is essential to support. Shall we do a followup to build this or there are some difficulties with it I do not understand?

t0yv0 (approved):

PR #327, opened 2025-05-17
Support converting local modules to resources parameterized by terraform-module

Zaid-Ajaj:
Following up #324 this time supporting local modules

Uses https://github.com/pulumi/pulumi/pull/19513 from pulumi v3.170.0+ that allows calculating the paths of the TF modules, relative to the output project directory

Updates the used terraform-module provider to v0.1.4 in package blocks which includes important fix https://github.com/pulumi/pulumi-terraform-module/pull/316 by @t0yv0 

 - [ ] Requires a CLI that has https://github.com/pulumi/pulumi/pull/19582 so that `pulumi convert` parameterized the packages using the right paths
 - [x] Maybe needs another pass on updated dependencies (pulumi-yaml, tfbridge, pu/pu)?


pkg/convert/tf.go
```diff
@@ -3856,16 +3911,20 @@ func getPackageBlock(name string, prov *configs.RequiredProvider) (*hclwrite.Blo
...
 	block := hclwrite.NewBlock("package", []string{packageName})
 	body := block.Body()
 	body.SetAttributeValue("baseProviderName", cty.StringVal("terraform-module"))
-	body.SetAttributeValue("baseProviderVersion", cty.StringVal("0.1.3"))
+	body.SetAttributeValue("baseProviderVersion", cty.StringVal("0.1.4"))
 	paramBlock := hclwrite.NewBlock("parameterization", []string{})
+	appendComment(body, "encoded parameterization values:")
+	appendComment(body, fmt.Sprintf("module: %s", source.String()))
+	appendComment(body, fmt.Sprintf("version: %s", version))
+	appendComment(body, fmt.Sprintf("packageName: %s", packageName))
```
Frassle, line 3927: I would have put these just above the base64 text

t0yv0:
Some discussion on https://github.com/pulumi/pulumi/pull/19582 that seems healthy, we need to work through the details there and figure it out.

In the meanwhile @Zaid-Ajaj is there something here independently useful? Without 19582, is there a prescriptive folder scheme we can recommend to users in a README that would still work?

pkg/convert/tf.go
Zaid-Ajaj: Fixed ✅

pkg/convert/tf.go
```diff
@@ -3234,6 +3249,21 @@ func translateModuleSourceCode(
...
 
+			if source, isLocal := moduleFromLocalFileSystem(moduleCall); isLocal {
+				key := filepath.Clean(filepath.Join(sourceDirectory, string(source)))
+				if _, seen := sandboxedModules[key]; !seen && wrapUsingTerraformModule {
+					state.sandboxedModuleNames[moduleCall.Name] = moduleName
+					sandboxedModules[key] = &PulumiTerraformModule{
+						packageName:     packageName,
+						moduleCall:      moduleCall,
+						local:           true,
+						absolutePath:    key,
```
t0yv0, line 3260: Curious is this truly absoultePath? filepath.Clean does not absolutize but perhaps sourceDirectory is absolute already.

pkg/convert/tf.go
```diff
@@ -3639,15 +3674,23 @@ func translateModuleSourceCode(
...
+					}
 
-				modulePackegeBlock := sandboxedModulePackageBlock(sandboxModule.packageName, source, version.String())
-				body.AppendBlock(modulePackegeBlock)
-				body.AppendUnstructuredTokens(hclwrite.TokensForIdentifier("\n"))
+					modulePackegeBlock := remotePulumiTerraformModulePackageBlock(sandboxModule.packageName, source, version.String())
+					body.AppendBlock(modulePackegeBlock)
+					body.AppendUnstructuredTokens(hclwrite.TokensForIdentifier("\n"))
+				} else {
+					relativePath, _ := filepath.Rel(sandboxModule.targetDirectory, sandboxModule.absolutePath)
```
t0yv0, line 3689: Handle the error here, perhaps `contract.NoErrorf`?

pkg/convert/tf.go
```diff
@@ -3639,15 +3674,23 @@ func translateModuleSourceCode(
...
+					modulePackageBlock := localPulumiTerraformModulePackageBlock(sandboxModule.packageName, relativePath)
+					body.AppendBlock(modulePackageBlock)
+					body.AppendUnstructuredTokens(hclwrite.TokensForIdentifier("\n"))
```
t0yv0, line 3692: Curious newline syntax 🤔

pkg/convert/tf.go
```diff
@@ -3880,9 +3939,51 @@ func sandboxedModulePackageBlock(
...
+	packageName string,
+	localPath string,
+) *hclwrite.Block {
+	block := hclwrite.NewBlock("package", []string{packageName})
+	body := block.Body()
+	body.SetAttributeValue("baseProviderName", cty.StringVal("terraform-module"))
+	body.SetAttributeValue("baseProviderVersion", cty.StringVal("0.1.4"))
+	paramBlock := hclwrite.NewBlock("parameterization", []string{})
+
+	// hardcoded version to align with the version in pulumi-terraform-module
```
t0yv0, line 3962: Ah interesting, thanks for this comment. Probably not worth build-depending on the pulumi-terraform-module code for this. Perhaps to make it more prominent we can lift it to a top-level const.

pkg/convert/tf.go
Zaid-Ajaj: Added an assert ✅

pkg/convert/tf.go
Zaid-Ajaj: I _believe_ it is absolute because `sourceDirectory` should be absolute. Could add more tests in a follow up

PR #334, opened 2025-06-18
Convert benchmark

VenelinMartinov:
Adds some code for running a TF conversion benchmark. The plan is to:

1. Get a collection of TF programs and verify that `tofu plan`, and `tofu apply` work fine, as well as some program-specific assertion (ex. my bucket is configured correctly, etc.)
2. Run `pulumi convert` on that set of programs and verify that `pulumi preview` and `pulumi up` work fine, as well as the specific assertion from above.
3. Run the same via some LLM convert and see what the results are.

The first two are present here as well as a simple program based on the `random` provider.

### DIFF
--- a/tests/benchmark/.gitignore
+++ b/tests/benchmark/.gitignore
@@ -1,0 +1,1 @@
+**/.terraform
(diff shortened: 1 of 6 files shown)


mikhailshilkov (approved):
Makes sense as a start! I got the first benchmark running locally.

VenelinMartinov:

- #335
    - #334 ◀

PR #336, opened 2025-06-18
AWS S3 bucket example for benchmark

VenelinMartinov:


tests/benchmark/tofu.go
```diff
@@ -88,6 +95,7 @@ func runTofuBenchmarks(testCases []testCase) map[string]*benchmarkResult {
 			results[tc.name].planSuccess = true
 		}
 
+		defer runTofuDestroy(dir)
```
mikhailshilkov, line 98: What's the practical difference between running destroy here vs where it used to be?

tests/benchmark/programs/aws_bucket/main.tf
```diff
@@ -0,0 +1,22 @@
+resource "random_string" "bucket_name" {
+  length  = 8
+  special = false
+  upper   = false
+}
+
+resource "aws_s3_bucket" "example" {
+  bucket = random_string.bucket_name.result
```
mikhailshilkov, line 8: Probably a nit and a stretch - but a human converter would probably drop the random resource from the TF program and use pulumi auto-naming for the bucket. Is there a benchmark for "produces idiomatic Pulumi code" opportunity for the future?

mikhailshilkov (approved):

PR #338, opened 2025-06-18
Add languages and better result display for convert benchmark

VenelinMartinov:
This adds a flag for languages and improves the result display of the convert benchmark.


mikhailshilkov:
An all-language run will take a very long time and will be slow to iterate on. We should probably have a flag to pass which languages to run. What do you think?

PR #340, opened 2025-06-23
Add AWS lambda example to conversion benchmark

VenelinMartinov:
This adds a non-trivial AWS lambda example to the conversion benchmark. Adapted from https://developer.hashicorp.com/terraform/tutorials/aws/lambda-api-gateway

PR #341, opened 2025-06-23
Add a bucket with tags example to conversion benchmark

VenelinMartinov:
Noticed that both the converter and the LLM struggle with provider config in some situations. The converter does not implement blocks in provider config and the LLM sometimes tries to generate some invalid YAML based on the provider config.

This PR adds a simple example to test that.

### DIFF
--- a/tests/benchmark/main.go
+++ b/tests/benchmark/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


VenelinMartinov:

- #340
    - #341 ◀

mikhailshilkov:
How are you thinking about many targeted examples vs. large examples testing multiple features? Could we add provider config to another existing AWS example?

VenelinMartinov:
superseded by https://github.com/pulumi/pulumi-converter-terraform/pull/345

PR #342, opened 2025-06-24
Extract the Claude prompt from the Pulumi MCP server

mikhailshilkov:
The Pulumi MCP server now contains a prompt that helps MCP hosts convert Terraform to Pulumi programs. This PR switches the Claude benchmark from a simplistic hard-coded prompt to the one from MCP.

One complication is that currently there is no way to tell Claude Code to use a given MCP prompt directly. Therefore, this PR goes in a roundabout way: it calls an MCP client to retrieve the prompt (parameterized with the folder name) and then passes this prompt to Claude Code.


tests/benchmark/pul.go
```diff
@@ -44,18 +49,50 @@ func runPulumiConvertYaml(srcDir string, outDir string) error {
...
 		return err
 	}
 
-	_, err = run(outDir, "pulumi", "install")
+	stdout, err := run(srcDir, "claude", "-p", prompt, "--add-dir", outDir, "--dangerously-skip-permissions")
+	fmt.Printf("Claude convert stdout: %s\n", stdout)
 	return err
 }
 
+func readMcpPrompt(mcpServer, promptName, outDir string) (string, error) {
```
VenelinMartinov, line 62: Hope they add something better for this 😄

VenelinMartinov (approved):

PR #348, opened 2025-06-25
Add cloudflare plan-only example

VenelinMartinov:
This adds a Cloudflare static website plan-only example. I had some trouble getting this to work end-to-end.

```
--------------------------------
tfResults:
total: 1
convertSuccesses: 1 (100%)
planSuccesses: 1 (100%)
planComparisonSuccesses: 1 (100%)
applySuccesses: 0 (100%)
assertSuccesses: 0 (100%)
claudeResults:
total: 1
convertSuccesses: 1 (100%)
planSuccesses: 1 (100%)
planComparisonSuccesses: 1 (100%)
applySuccesses: 0 (100%)
assertSuccesses: 0 (100%)
pulumiResultsTs:
total: 1
convertSuccesses: 1 (100%)
planSuccesses: 1 (100%)
planComparisonSuccesses: 1 (100%)
applySuccesses: 0 (100%)
assertSuccesses: 0 (100%)
```


mikhailshilkov (approved):
LGTM.

A comment to the other PR: when printing results, we should probably adjust the output on whether the test is plan-only and don't print unrelated metrics.

VenelinMartinov:
The result here is from a single example but the summary is intended to run for all examples - some of them are not plan-only so the metrics should still be relevant.

The intention is that all examples should be the default way to run this and we are correctly adjusting the denominators on all the metrics depending on the total number of relevant examples.

PR #350, opened 2025-06-25
Benchmark add EKS example

VenelinMartinov:
This PR adds a more involved EKS example. It uses modules and the hope is that when we convert the program we should use the module provider for the module references. The example is added as a plan-only test since EKS clusters take >10 minutes to provision which makes this way too painful.

Both the converter and the LLM conversion struggle here. The converter fails with:

```
Pulumi Version:   v3.175.0
error: main.pp:95,1-129,2: circular reference; 
error: main.pp:139,63-76: unknown property 'cluster_name' among [id urn]; 
error: main.pp:140,38-52: unknown property 'oidc_provider' among [urn id]; 
error: Detected that /Users/vvm/.pulumi/plugins/resource-terraform-module-v0.1.4/pulumi-resource-terraform-module exited prematurely. 
       This is *always* a bug in the provider. Please report the issue to the provider author as appropriate.
       To assist with debugging we have dumped the STDOUT and STDERR streams of the plugin:

error: could not generate output program: error generating packages: generated root is not a valid pulumi workspace "/var/folders/82/nqnqw81s1h56l5nv940f9mq00000gn/T/pulumi-benchmark3942479593": no Pulumi.yaml project file found (searching upwards from /private/var/folders/82/nqnqw81s1h56l5nv940f9mq00000gn/T/pulumi-benchmark3942479593). If you have not created a project yet, use `pulumi new` to do so: no project file found
```

The LLM generated a program but it had too few resources - I don't think it has access to the remote module source files so it tries to invent what they do.

```
--------------------------------
tfResults:
total: 1
convertSuccesses: 1 (100%)
planSuccesses: 1 (100%)
planComparisonSuccesses: 1 (100%)
applySuccesses: 0 (100%)
assertSuccesses: 0 (100%)
claudeResults:
total: 1
convertSuccesses: 1 (100%)
planSuccesses: 1 (100%)
planComparisonSuccesses: 0 (0%)
applySuccesses: 0 (100%)
assertSuccesses: 0 (100%)
pulumiResultsTs:
total: 1
convertSuccesses: 0 (0%)
planSuccesses: 0 (0%)
planComparisonSuccesses: 0 (0%)
applySuccesses: 0 (100%)
assertSuccesses: 0 (100%)
```

### DIFF
--- a/tests/benchmark/plans/aws_eks_cluster/pulumi_plan_converter-ts.out.json
+++ b/tests/benchmark/plans/aws_eks_cluster/pulumi_plan_converter-ts.out.json
@@ -1,0 +1,164 @@
+{
+  "resourcePlans": {
+    "urn:pulumi:test::aws_eks_cluster::pulumi:providers:aws::default_6_83_0": {
+      "steps": [
+        "create"
+      ]
+    },
+    "urn:pulumi:test::aws_eks_cluster::pulumi:providers:pulumi::default": {
+      "steps": [
+        "create"
+      ]
+    },
+    "urn:pulumi:test::aws_eks_cluster::pulumi:providers:random::default_4_18_2": {
+      "steps": [
+        "create"
+      ]
+    },
+    "urn:pulumi:test::aws_eks_cluster::pulumi:providers:std::default_2_2_0": {
+      "steps": [
+        "create"
+      ]
+    },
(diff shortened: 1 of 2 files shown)


VenelinMartinov:

- #350 ◀

mikhailshilkov:
Nice! Any chance you converted the program or any module usage manually? If not, I can do that and then maybe see how I can explain it to LLMs.

mikhailshilkov (approved):

PR #359, opened 2025-07-07
Update pu/pu version and bridge

VenelinMartinov:
We are using outdated pu/pu and bridge versions which was causing some test failures.

fixes https://github.com/pulumi/pulumi-converter-terraform/issues/357
fixes https://github.com/pulumi/pulumi-converter-terraform/issues/358


t0yv0 (commented):
Curious if we know why the tests are failing. 

Also can we automate this with Renovate as this is presumably a recurring task?

t0yv0 (approved):

VenelinMartinov:
Note the converter also depends on our plugin-sdk fork, so this is not quite trivial

PR #389, opened 2026-01-13
Fix camelCase conversion for string-indexed object properties

### LINKED ISSUE #373: Indexing an object via a string key doesn't use the camelCased version of the key
### What happened?

When indexing an object via a string key (`object["my_value"]`) the converter code uses the key as is instead of using the camelCase version `object["myValue"]` 

### Example

This is correct ✅ 
```tf
locals {
  accounts = {
    "us-east-1" = { aws_region = "us-east-1", something = 10 },
    "us-west-2" = { aws_region = "us-west-2", something = 20 },
    "eu-west-1" = { aws_region = "eu-west-1", something = 30 }
  }
}

resource "random_string" "name" {
  for_each = local.accounts
  length   = each.value.aws_region
}
```
Translates to:
```
accounts = {
  "us-east-1" = {
    awsRegion = "us-east-1"
    something = 10
  }
  "us-west-2" = {
    awsRegion = "us-west-2"
    something = 20
  }
  "eu-west-1" = {
    awsRegion = "eu-west-1"
    something = 30
  }
}

resource "name" "random:index/randomString:RandomString" {
  options {
    range = accounts
  }
  length = range.value.awsRegion
}
```
This is incorrect ❌ 
```tf
resource "random_string" "name" {
  for_each = local.accounts
  length   = each.value["aws_region"]
}
```
Translates to
```
resource "name" "random:index/randomString:RandomString" {
  options {
    range = accounts
  }
  length = range.value["aws_region"]
}
```
PCL errors:
```
Pulumi Version:   v3.189.0
error: main.pp:19,23-37: unknown property 'aws_region' among [something awsRegion]; 
error: main.pp:19,23-37: unknown property 'aws_region' among [awsRegion something]; 
error: main.pp:19,23-37: unknown property 'aws_region' among [awsRegion something]; 
error: main.pp:19,23-37: cannot traverse value of type union(object({awsRegion = string, something = cty.NumberIntVal(10)}), object({awsRegion = string, something = cty.NumberIntVal(20)}), object({awsRegion = string, something = cty.NumberIntVal(30)})); 
error: could not generate output program
```

### Output of `pulumi about`

Pulumi CLI v3.189.0
Converter v1.2.3

### Additional context

_No response_

### Contributing

joeduffy:
When indexing an object via a string key using bracket notation like
`each.value["aws_region"]`, the converter now properly applies camelCase
transformation to the key, converting it to `range.value["awsRegion"]`.

Previously, string keys in bracket notation were passed through unchanged,
causing PCL type validation errors since the schema only recognizes
camelCased property names.

The fix applies camelCase transformation in two places:
- `rewriteRelativeTraversal()` for traversal-based index access
- `convertIndexExpr()` for standalone index expressions

Fixes #373

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -3,3 +3,5 @@
  - Handle implicitly required providers that are non-Pulumi bridged providers by resolving and parameterizing them using their latest version from the registry.
 
 ### Bug Fixes
+
+ - Fix camelCase conversion for string-indexed object properties. Bracket notation like `each.value["aws_region"]` now correctly converts to `range.value["awsRegion"]`.


Frassle (commented):
Does this work for object variables as well? Can think of two cases there.
1. config variables where the user config might set that the object key is camel case and we can't see that from the program.
2. local variable set via an object expression, I can't recall if we camelCase all object expressions.

PR #413, opened 2026-04-08
`try` is supported is PCL

### LINKED ISSUE #16: Implement try function
`try` is one of the most common terraform functions that is now emitted as `notImplemented("try(...)")` implementing this would be great! It must be implemented as a PCL function intrinsic `try(...)` accepting any number of arguments `try(arg1, arg2, ..., argN)`

My proposal is for each code generator, it should convert the expressions into lazy expressions, then return the first argument that does not error out.

for example from PCL -> C#
```
try(subnets[0].id, config.subnetId)
```
generates 
```cs
Try(() => subnets[0].Id, () => config.subnetId)
```

iwahbe:
Fixes #16


pkg/convert/testdata/programs/builtin_functions/main.tf
```diff
@@ -1184,21 +1184,6 @@ output "funcTrimsuffix2" {
...
-}
-output "funcTry1" {
-  value = try(local.foo.bar, "fallback")
-}
-output "funcTry2" {
-  value = try(local.foo.boop, "fallback")
-}
-output "funcTry3" {
-  value = try(local.nonexist, "fallback")
-}
```
iwahbe, line 1199: This program was not valid TF, since `local.nonexist` doesn't exist, and TF errors even with `try`:

```
        ╷
        │ Error: Reference to undeclared local value
        │ 
        │   on main.tf line 3, in output "should_fail":
        │    3:   value = try(local.missing, "never")
        │ 
        │ A local value with the name "missing" has not been declared.
        ╵
```

scripts/generate_builtins.py
```diff
@@ -19,6 +19,11 @@ def trimext(file: str) -> str:
     return file[0:i]
 
 # Some functions don't have any examples in the Terraform docs. They _must_ have overrides here else we error on generation.
```
Frassle: Move this comment down to the overrides block

scripts/generate_builtins.py
```diff
@@ -18,7 +18,12 @@ def trimext(file: str) -> str:
         return file
     return file[0:i]
 
+skip = {"try"}
+
 # Some functions don't have any examples in the Terraform docs. They _must_ have overrides here else we error on generation.
+# Functions that are tested via conformance tests rather than this generated test.
+# These are excluded because their TF doc examples use intentionally invalid
+# expressions (e.g. accessing non-existent properties) that PCL's binder rejects.
```
Frassle:
```suggestion
# Functions that are tested via conformance tests rather than this generated test.
# These are excluded because their TF doc examples use intentionally invalid
# expressions (e.g. accessing non-existent properties) that PCL's binder rejects.
skip = {"try"}

# Some functions don't have any examples in the Terraform docs. They _must_ have overrides here else we error on generation.
```

Frassle (approved):

PR #415, opened 2026-04-09
Generate `list(optional(T))` when the default value is  `[null]`

### LINKED ISSUE #398: pulumi convert fails when converting Databricks Security Reference Architecture
### What happened?

I'm trying to convert the Terraform code based on https://databricks.github.io/terraform-databricks-sra/docs/usage/AWS/gettingstarted/

It failed with the following error:

```
Converting to python...
Downloading provider: std
================================================================================
The Pulumi CLI encountered a code generation error. This is a bug!
We would appreciate a report: https://github.com/pulumi/pulumi/issues/
Please provide all of the below text in your report.
================================================================================
Pulumi Version:   v3.218.0
error: variables.pp:116,17-23: cannot assign expression of type (null) to location of type list(output(string) | string) | output(list(string)): ; 
error: variables.pp:122,17-23: cannot assign expression of type (null) to location of type list(output(string) | string) | output(list(string)): ; 
error: variables.pp:340,17-23: cannot assign expression of type (null) to location of type list(output(string) | string) | output(list(string)): ; 
```

I was able to workaround it with help from Claude and a few edits.
```
diff --git a/aws/tf/variables.tf b/aws/tf/variables.tf
index 0b757c8..98d7381 100644
--- a/aws/tf/variables.tf
+++ b/aws/tf/variables.tf
@@ -161,14 +161,14 @@ variable "private_subnets_cidr" {
   description = "CIDR blocks for private subnets."
   type        = list(string)
   nullable    = true
-  default     = [null]
+  default     = null
 }
 
 variable "privatelink_subnets_cidr" {
   description = "CIDR blocks for private link subnets."
   type        = list(string)
   nullable    = true
-  default     = [null]
+  default     = null
 }
 
 variable "region" {
@@ -342,7 +342,7 @@ variable "sg_egress_ports" {
   description = "List of egress ports for security groups."
   type        = list(string)
   nullable    = true
-  default     = [null]
+  default     = null
 }
```

There were also a lot of warnings about "Function try not yet implemented" - that would make this conversion a lot smoother!

### Example

git clone git@github.com:databricks/terraform-databricks-sra.git
cd terraform-databricks-sra/aws/tf
pulumi convert --from terraform --language python --out pulumi

### Output of `pulumi about`

$ pulumi about
CLI          
Version      3.218.0
Go Version   go1.25.6
Go Compiler  gc

Host     
OS       darwin
Version  26.2
Arch     arm64

Backend        
Name           <Redacted>
URL            <Redacted - on S3>
User           <Redacted>
Organizations  
Token type     personal

Pulumi locates its logs in /var/folders/lp/... by default

### Additional context

_No response_

### Contributing

iwahbe:
This assumes that `optional` is a valid type constructor in PCL, which currently it is not.

Fixes #398


pkg/convert/tf.go
```diff
@@ -2351,6 +2351,25 @@ func convertVariable(state *convertState, scopes *scopes,
 		pulumiType = inferPrimitiveType(variable.Default.Type(), pulumiType)
 	}
 
+	// If the default value contains null elements in a list or set, wrap the element
```
Frassle: Should we just always wrap with optional? Can you set null via the config system? Even if not this also doubles up for modules and I imagine you can pass null to those?

iwahbe: This should have been in draft. I need to explore if how PCL handles nested null conversions & type errors. Right now (without https://github.com/pulumi/pulumi/pull/22547), this isn't even expressible in PCL.

Frassle:
I think my comment still stands. Isn't it more correct to always type as optional because you could pass null via the config or module system and we can't tell from the TF source if it should be nullable or not?

pkg/convert/tf.go
iwahbe: https://github.com/pulumi/pulumi/pull/22547 has merged, this is now expressible.

> Should we just always wrap with optional?

That would be more correct at runtime, in exchange for introducing noise to our generated files. I'm not sure where the correct balance is here, especially since we mostly use this conversion for example generation.

> Can you set null via the config system?

Yes. 

> Even if not this also doubles up for modules and I imagine you can pass null to those?

Yes.

iwahbe:
> I think my comment still stands. Isn't it more correct to always type as optional because you could pass null via the config or module system and we can't tell from the TF source if it should be nullable or not?

As I said in https://github.com/pulumi/pulumi-converter-terraform/pull/415#discussion_r3672904089:

> That would be more correct at runtime, in exchange for introducing noise to our generated files. I'm not sure where the correct balance is here, especially since we mostly use this conversion for example generation.

Frassle (approved):
I think we take correctness here, sad it adds so many optionals :(

PR #427, opened 2026-04-23
Implement specialized conversion for helm_release resources

guineveresaenger:
## Summary

- Adds a specialized conversion path for Terraform `helm_release` resources, translating `set {}` / `set_list {}` / `set_sensitive {}` blocks into a single `values` object (with `secret()` applied to sensitive values) and mapping `repository*` fields onto `repositoryOpts`.
- Includes a `partial_helm_release` test program with golden PCL and diagnostics.
- Rebased onto current `main` and updated to match the current `scopes.pulumiName` and `maps.Keys` APIs.

Original work by @Zaid-Ajaj in #198 (closed because the branch was deleted); this PR revives and rebases that work. Credit goes to Zaid for the implementation.

## Test plan

- [x] `go test ./pkg/convert/...` passes
- [x] `go test ./tests/...` passes (with pulumi-aws 7.x installed locally)
- [x] `go test ./tests/conformance/...` passes
- [x] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pkg/convert/testdata/programs/partial_helm_release/pcl/main.pp
```diff
@@ -0,0 +1,49 @@
+package "helm" {
+  baseProviderName    = "terraform-provider"
+  baseProviderVersion = "0.8.1"
+  parameterization {
+    version = "3.1.1"
+    name    = "helm"
+    value   = "eyJyZW1vdGUiOnsidXJsIjoicmVnaXN0cnkudGVycmFmb3JtLmlvL2hhc2hpY29ycC9oZWxtIiwidmVyc2lvbiI6IjMuMS4xIn19"
+  }
+}
```
iwahbe: This shouldn't be here, since there is only one resource in the file and it's for the `kubernetes` provider.

pkg/convert/tf.go
```diff
@@ -2097,6 +2112,152 @@ func bodyContent(body hcl.Body) *hcl.BodyContent {
 	return content
 }
 
+// A specialized version of convertBody that is specific to converting the body of helm_release resources
+func convertHelmReleaseResource(state *convertState, scopes *scopes, fullyQualifiedPath string, body hcl.Body) bodyAttrsTokens {
```
iwahbe: Can you help me understand what part of this can't be expressed with the normal `GetMapping` machinery?

`convertBody` is a non-trivial function. Would it be possible to use the existing `convertBody` and instead have the helm release customizer function pre-process the resource so it's the right shape for the normal conversion?

pkg/convert/tf.go
```diff
@@ -2735,7 +2899,15 @@ See https://www.pulumi.com/docs/iac/concepts/options/deletebeforereplace/ for de
...
 
-	resourceArgs := convertBody(state, scopes, path, managedResource.Config)
+	var resourceArgs bodyAttrsTokens
+	if _, custom := customResourceMappings[managedResource.Type]; custom {
+		if managedResource.Type == "helm_release" {
+			resourceArgs = convertHelmReleaseResource(state, scopes, path, managedResource.Config)
+		}
+	} else {
+		resourceArgs = convertBody(state, scopes, path, managedResource.Config)
+	}
```
iwahbe: Can we push `convertHelmReleaseResource` into `customResourceMappings` so we tie a successful lookup to matching a convert function, and move everything help specific into one place.

```suggestion
	if _, custom := customResourceMappings[managedResource.Type]; custom {
		resourceArgs = custom.convert(state, scopes, path, managedResource.Config)
	} else {
		resourceArgs = convertBody(state, scopes, path, managedResource.Config)
	}
```

pkg/convert/tf.go
```diff
@@ -3268,7 +3440,7 @@ func translateModuleSourceCode(
 			}
 
 			providerInfo, err := info.GetProviderInfo(provider, requiredProviders[provider])
-			if err != nil {
+			if err != nil && !isCustomProviderMapping(provider) {
```
iwahbe: This feels wrong, especially when we might be trying a lookup for a resource that is not the custom resource.

pkg/convert/tf_state.go
```diff
@@ -224,7 +224,7 @@ func TranslateState(info ProviderInfoSource, path string) (*plugin.ConvertStateR
 					tfType := resource.Addr.Resource.Type
 					provider := impliedProvider(tfType)
 					providerInfo, err := info.GetProviderInfo(provider, nil /*requiredProvider*/)
-					if err != nil {
+					if err != nil && !isCustomProviderMapping(provider) {
```
iwahbe: Likewise https://github.com/pulumi/pulumi-converter-terraform/pull/427/changes#r3129787000

pkg/convert/testdata/programs/partial_helm_release/pcl/diagnostics.json
iwahbe, line 1: These are all user visible warnings. If the example has a correct TF resource and that maps correctly into a Pulumi resource, we shouldn't see any warnings.

pkg/convert/tf.go
Frassle: Yeh I would much rather get this working via GetMapping even if that means adding some more machinery to that. Special casing this just for helm feels horrible for maintenance.

pkg/convert/tf.go
guineveresaenger: Moved to `CustomResourceMapping`.

pkg/convert/testdata/programs/partial_helm_release/pcl/diagnostics.json
guineveresaenger: I've implemented the remaining logic to get rid of all warnings save one. There isn't a `pass_credentials` in our helm so the warning seems fair (rather than a silent drop)

pkg/convert/tf.go
guineveresaenger: This conversion is a very special use case: a convert from a TF provider to a non-TF native resource.
Currently GetMapping can't handle differently nested shapes, which is only one of the differences; there are some fields different enough that we'd still need pre-processing.
Given that the TF -> native conversion is a very unusual ask I didn't think it was worth the effort to generalize GetMapping, and instead propose this one-off. If we ever get similarly shaped requests (pun intended) we can re-evaluate.

I've refactored the function to only concern itself with processing anything GetMapping can't handle.

pkg/convert/tf_helm_test.go
```diff
@@ -0,0 +1,364 @@
...
+		expr := parseExprForTest(t, `[<<EOT
+a: 1
+b: two
+EOT
+]`)
+		tuple := expr.(*hclsyntax.TupleConsExpr)
+		got, ok := mergeStaticHelmValues(tuple)
+		require.True(t, ok)
+		assert.True(t, got["a"].RawEquals(cty.NumberIntVal(1)))
+		assert.True(t, got["b"].RawEquals(cty.StringVal("two")))
```
iwahbe: Does `assert.Equals(t, ..., got)` work? If so, can we do that? Likewise for the other tests that use lots of `assert.True(t, x.RawEquals(y))`.

pkg/convert/tf.go
```diff
@@ -2097,6 +2121,373 @@ func bodyContent(body hcl.Body) *hcl.BodyContent {
 	return content
 }
 
+// helm_release blocks whose name/value pairs collapse into a single `values` map.
```
iwahbe: Can we quarantine all the helm specific stuff to it'd own file: `pkg/convert/helm.go`?

iwahbe (approved):
I'm trusting that you've verified that this produces a valid helm resource for Pulumi.

I don't love the special casing, but I do think it's the fastest way to ship this and it doesn't force much complexity on the rest of the system.

pkg/convert/tf.go
iwahbe: At minimum, something must go in the converter to say: "see that helm resource, ask the kubernetes plugin about it".

PR #432, opened 2026-04-25
Convert `remote-exec` provisioners

### LINKED ISSUE #430: Convert `remote-exec` provisioner to `command:remote:Command`
## Hello!

## Issue details

We should convert the `remote-exec` provisioner to `command:remote:Command`, mirroring how we convert `local-exec` to `command:local:Command`. Currently, we just emit a comment explaining that the feature isn't supported in conversion.

iwahbe:
This PR adds the ability to convert `remote-exec` provisioners into `command:remote:Command` resources. This mirrors how we handle `local-exec` -> `command:local:Command`. `remote-exec` has 3 ways to pass a set of commands to run:

- `inline`: where we just create a `command:remote:Command`
- `script`: where we upload a file with `command:remote:CopyToRemote`
- `scripts`: where we upload a set of files with `command:remote:CopyToRemote` and `range`, then exec all of them.

Testing is handled by an in-container remote server.

Fixes https://github.com/pulumi/pulumi-converter-terraform/issues/430


pkg/convert/testdata/programs/remote_exec_provisioners/pcl/diagnostics.json
```diff
@@ -0,0 +1,4 @@
+[
+  "warning:remote_exec_provisioners/main.tf:30,5-29:Unsupported connection attribute:Connection attribute \"agent\" has no equivalent on the Pulumi command:remote:Connection type and was dropped.",
+  "warning:remote_exec_provisioners/main.tf:29,5-29:Unsupported connection attribute:Connection attribute \"timeout\" has no equivalent on the Pulumi command:remote:Connection type and was dropped."
```
Frassle: Couldn't we copy this to create/update timeouts?

Frassle (approved):

pkg/convert/testdata/programs/remote_exec_provisioners/pcl/diagnostics.json
iwahbe: I think so.

PR #441, opened 2026-05-11
Convert TF for_each idioms to string-keyed PCL ranges

### LINKED ISSUE #228: Attempting conversion containing `for_each` with primitives fails

iwahbe:
## Summary

Fixes #228 by detecting two TF `for_each` idioms during conversion and rewriting each to a PCL form the binder can handle.

- `for_each = toset(<list>)` on a resource → `range = { for entry in <list> : entry => entry }`. Without this the converter emits `range = invoke(\"std:index:toset\", ...).result`, which types as a list/set; PCL then requires integer indexing, so `name[\"alpha\"]` fails to bind ("a number is required").
- `for_each = <cond> ? [<x>] : []` on a `dynamic` block (detected only when both branches are tuple literals, truthy has exactly one element, falsy is empty) → `<cond> ? [<content>] : []` with `each.value`/`each.key` inlined as `<x>`/`0` in the content. The naive emit was a for-expression over a tuple-of-number whose iterator value types as a union of `0` and `1`, which is what the bug report was hitting. The same idiom is invalid TF at the resource level (TF requires for_each to be a map or set of strings), so no rewrite is applied there.

## Test plan

- [x] `TestL2ForEachStringKey` (previously skipped) now passes — golden PCL is the for-object form, both TF apply and Pulumi up succeed and produce equal outputs.
- [x] New `TestL2DynamicBlockConditionalGate` exercises the gate idiom in a dynamic block, referencing `rule.value` in the content so the inlining path is covered. Golden shows the literal substituted in.
- [x] New `pkg/convert/testdata/programs/for_each_toset` and `dynamic_block_conditional_gate` translate-snapshot programs and unit tests for the new detection helpers, lifting patch coverage above the codecov threshold.
- [x] `go test ./...` is green locally.


pkg/convert/testdata/programs/for_each_toset/pcl/main.pp
```diff
@@ -0,0 +1,11 @@
+resource "tagged" "simple:index:resource" {
+  options {
+    range = { for entry in ["alpha", "beta"] : entry => entry }
```
Frassle, line 3: Isn't this translation wrong if the list has duplicate?

iwahbe: No, because the source material on the `for_each` is `toset(["alpha", "beta"])`. pkg/convert/testdata/programs/for_each_toset/main.tf is what we are converting from.

Frassle: Yeh but what if the source code was `toset(["alpha", "alpha", "beta"])`?

iwahbe: Then the resource would correctly range over the value `{"alpha": "alpha", "beta": "beta"}`, because the `for _ in .. : k => v` resolves to a map which will de-dupe keys.

Frassle: Ah I see, subtle

Frassle (approved):

